PR #23, opened 2023-10-18
chore(cloudflare): add option to set persistence dir

alexanderniebuhr:
## Changes

- Fixes a DX issue with Cloudflare runtime in `astro dev`
- Now you should be able to use `wrangler` as you are used to, to seed db or apply migrations
- **Open question:** see code comments of self-review
- Don't forget a changeset! `pnpm exec changeset`

## Testing

- existing test

## Docs

- cc @withastro/maintainers-docs

### DIFF
--- a/.changeset/rich-foxes-deliver.md
+++ b/.changeset/rich-foxes-deliver.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Adds a new property `persistTo` which allows setting the directory for local state files when using Cloudflare runtime with `astro dev`. This is useful when you want to persist state between restarts of the dev server, for example when using KV, D1, R2 to store data.
+
+Additionally, updates the format of the `runtime` configuration and adds a warning when the deprecated format is used. The current format is now `runtime: { mode: 'off' | 'local', persistTo: string }`. See [runtime documentation](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime) for more information.
--- a/packages/cloudflare/README.md
+++ b/packages/cloudflare/README.md
@@ -186,16 +186,24 @@ export default defineConfig({
 
 ### `runtime`
 
-`runtime: "off" | "local"`
+`runtime: { mode: "off" | "local", persistTo: string }`
 
-default `"off"`
+default `{ mode: 'off', persistTo: '' }`
 
 Determines whether and how the Cloudflare Runtime is added to `astro dev`.
 
 The Cloudflare Runtime includes [Cloudflare bindings](https://developers.cloudflare.com/pages/platform/functions/bindings), [environment variables](https://developers.cloudflare.com/pages/platform/functions/bindings/#environment-variables), and the [cf object](https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties). Read more about [accessing the Cloudflare Runtime](#cloudflare-runtime).
 
+The `mode` property defines how the runtime is added to `astro dev`:
+
 - `local`: uses bindings mocking and locally static placeholders
 - `off`: no access to the Cloudflare runtime using `astro dev`. You can alternatively use [Preview with Wrangler](#preview-with-wrangler)
+
+The `persistTo` property defines where the local runtime is persisted to when using `mode: local`. This value is a directory relative to your `astro dev` execution path.
+
+The default value is set to `.wrangler/state/v3` to match the default path Wrangler uses. This means both tools are able to access and use the local state.
+
+Whichever directory is set in `persistTo`, `.wrangler` or your custom value, must be added to `.gitignore`.
 
 ```diff lang="js"
 // astro.config.mjs
@@ -205,7 +213,7 @@ import cloudflare from '@astrojs/cloudflare';
 export default defineConfig({
   output: 'server',
   adapter: cloudflare({
-+   runtime: 'local',
++   runtime: { mode: 'local' },
   }),
 });
 ```
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -26,6 +26,7 @@ import { wasmModuleLoader } from './utils/wasm-module-loader.js';
 export type { AdvancedRuntime } from './entrypoints/server.advanced.js';
 export type { DirectoryRuntime } from './entrypoints/server.directory.js';
 
+type CF_RUNTIME = { mode: 'off' } | { mode: 'remote' } | { mode: 'local'; persistTo: string };
 type Options = {
 	mode?: 'directory' | 'advanced';
 	functionPerRoute?: boolean;
@@ -43,11 +44,18 @@ type Options = {
 		exclude?: string[];
 	};
 	/**
+	 * Going forward only the object API should be used. The modes work as known before:
 	 * 'off': current behaviour (wrangler is needed)
 	 * 'local': use a static req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 * 'remote': use a dynamic real-live req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 */
-	runtime?: 'off' | 'local' | 'remote';
+	runtime?:
+		| 'off'
+		| 'local'
+		| 'remote'
+		| { mode: 'off' }
+		| { mode: 'remote' }
+		| { mode: 'local'; persistTo?: string };
 	wasmModuleImports?: boolean;
 };
 
@@ -59,6 +67,14 @@ interface BuildConfig {
 	split?: boolean;
 }
 
+const RUNTIME_WARNING = `You are using a deprecated string format for the \`runtime\` API. Please update to the current format for better support. 
+
+Example:
+Old Format: 'local'
+Current Format: { mode: 'local', persistTo: '.wrangler/state/v3' }
+
+Please refer to our runtime documentation for more details on the format. https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime`;
+
 export default function createIntegration(args?: Options): AstroIntegration {
 	let _config: AstroConfig;
 	let _buildConfig: BuildConfig;
@@ -69,7 +85,21 @@ export default function createIntegration(args?: Options): AstroIntegration {
 
 	const isModeDirectory = args?.mode === 'directory';
 	const functionPerRoute = args?.functionPerRoute ?? false;
-	const runtimeMode = args?.runtime ?? 'off';
+
+	let runtimeMode: CF_RUNTIME = { mode: 'off' };
+	if (args?.runtime === 'remote') {
+		runtimeMode = { mode: 'remote' };
+	} else if (args?.runtime === 'local') {
+		runtimeMode = { mode: 'local', persistTo: '.wrangler/state/v3' };
+	} else if (
+		typeof args?.runtime === 'object' &&
+		args?.runtime.mode === 'local' &&
+		args.runtime.persistTo === undefined
+	) {
+		runtimeMode = { mode: 'local', persistTo: '.wrangler/state/v3' };
+	} else {
+		runtimeMode = args?.runtime as CF_RUNTIME;
+	}
 
 	return {
 		name: '@astrojs/cloudflare',
@@ -93,7 +123,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 					},
 				});
 			},
-			'astro:config:done': ({ setAdapter, config }) => {
+			'astro:config:done': ({ setAdapter, config, logger }) => {
 				setAdapter(getAdapter({ isModeDirectory, functionPerRoute }));
 				_config = config;
 				_buildConfig = config.build;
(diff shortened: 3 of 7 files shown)


alexanderniebuhr:
* main
  * **PR #23** <a href="https://app.graphite.dev/github/pr/withastro/adapters/23" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/withastro/adapters/23?utm_source=stack-comment).

packages/cloudflare/src/index.ts
```diff
@@ -47,7 +47,7 @@ type Options = {
 	 * 'local': use a static req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 * 'remote': use a dynamic real-live req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 */
-	runtime?: 'off' | 'local' | 'remote';
+	runtime?: 'off' | 'local' | 'remote' | { mode: 'off' | 'local' | 'remote'; persistTo: string };
```
alexanderniebuhr, line 50: Wondering how we could solve this. We want to use new API `{ mode: 'off' | 'local' | 'remote'; persistTo: string }` going forward, but still want to keep backwards compatibility, otherwise it might be a breaking change.

packages/cloudflare/src/index.ts
```diff
@@ -69,7 +69,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 	const functionPerRoute = args?.functionPerRoute ?? false;
-	const runtimeMode = args?.runtime ?? 'off';
+
+	let runtimeMode: { mode: 'off' | 'local' | 'remote'; persistTo: string } = {
+		mode: 'off',
+		persistTo: '',
+	};
+	if (typeof args?.runtime === 'object') runtimeMode = args?.runtime;
+	if (args?.runtime === 'local') runtimeMode = { mode: 'local', persistTo: './wrangler/state/v3' };
+	if (args?.runtime === 'remote') runtimeMode = { mode: 'remote', persistTo: '' };
```
alexanderniebuhr, line 80: This is all needed for backwards compatibility, and it does not even catch all edge-cases. I would like to outsource something like this to a lib, any suggestion for one, or how to handle it cleaner?

packages/cloudflare/src/index.ts
```diff
@@ -109,12 +137,17 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 					);
 				}
+
+				if (typeof args?.runtime === 'string') {
+					logger.warn(RUNTIME_WARNING);
+				}
 			},
 			'astro:server:setup': ({ server }) => {
-				if (runtimeMode !== 'off') {
+				console.log(runtimeMode.mode);
```
martrapp, line 146: Is it intended to leave this logging in the code?

packages/cloudflare/src/index.ts
```diff
@@ -47,7 +52,7 @@ type Options = {
 	 * 'local': use a static req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 * 'remote': use a dynamic real-live req.cf object, and env vars defined in wrangler.toml & .dev.vars (astro dev is enough)
 	 */
```
martrapp, line 54: The comment could mention that we are currently restructuring and want to support only the option object in the future.

martrapp (commented):
LGTM, only few questions.

packages/cloudflare/README.md
```diff
@@ -92,6 +92,7 @@ There are three options available:
   Example: For `src/pages/index.astro` (static), `src/pages/company.astro` (static), `src/pages/users/faq.astro` (static) and `/src/pages/users/[id].astro` (SSR) this will produce the following `_routes.json`:
 
   ```json
+  // dist/_routes.json
```
ematipico, line 95: We should remove it because comments aren't valid in JSON, and update the file name at line 92, which seems to be wrong.

packages/cloudflare/README.md
```diff
@@ -186,9 +188,9 @@ export default defineConfig({
 
 ### `runtime`
 
-`runtime: "off" | "local"`
+`runtime: "off" | "local" | { mode: "off" | "local", persistTo: string }`
 
-default `"off"`
+default `{ mode: 'off', persistTo: '' }`
```
ematipico, line 193: - We should document what's `persistTo` and explain to the user how it works.
- Also, it seems that `"off"` is basically `{ mode: "off", persistTo: '' }`, and we deprecated it. The documentation should mention it.
- `remote` is missing.

ematipico (commented):
The documentation requires some critical work before having this PR merged.

packages/cloudflare/README.md
alexanderniebuhr: Actually you are right. I was changing unrelated things in the docs. I'll tackle those `_routes.json` improvements in a separate PR & will focus for the docs only to the PR scope.

packages/cloudflare/README.md
alexanderniebuhr: Agreed.

We purposely chose to not include `remote` since it is not yet fully implemented.

matthewp (approved):

packages/cloudflare/src/index.ts
```diff
@@ -26,6 +26,11 @@ import { wasmModuleLoader } from './utils/wasm-module-loader.js';
 export type { AdvancedRuntime } from './entrypoints/server.advanced.js';
 export type { DirectoryRuntime } from './entrypoints/server.directory.js';
 
+interface CfRuntimeOptions {
+	mode: 'off' | 'local' | 'remote';
+	persistTo: string;
+}
```
ematipico, line 32: A suggestion. If `persistTo` is relevant only when using `local`, maybe we make types better:

```ts
type CfRuntimeOptions = 
	{ mode: 'off' } |
	{ mode: 'remote' } |
	{ mode: 'local', persistTo: string }

```

ematipico (approved):

alexanderniebuhr:
Just for context. I'm unhappy with `persistTo`, I just came up with it, because `wrangler` uses `--persist-to`, and `'persist-to'` is bad property name. So question should be try to be aligned with `wrangler`, so users know the setting does the same?

ematipico:
> Just for context. I'm unhappy with `persistTo`, I just came up with it, because `wrangler` uses `--persist-to`, and `'persist-to'` is bad property name. So question should be try to be aligned with `wrangler`, so users know the setting does the same?

I think you made a fantastic choice. It aligns with the CLI argument. It's only aligned with the naming convention (camel case) of our code. Plus, you can also document that `persistTo` is a direct implementation of the CLI.

packages/cloudflare/src/index.ts
ematipico: Yeah don't overthink, it's just a suggestion :) as long as the code works, that's what matters

packages/cloudflare/README.md
```diff
@@ -186,17 +186,21 @@ export default defineConfig({
...
 
+The `mode` property defines how the runtime is added to `astro dev`:
+
 - `local`: uses bindings mocking and locally static placeholders
 - `off`: no access to the Cloudflare runtime using `astro dev`. You can alternatively use [Preview with Wrangler](#preview-with-wrangler)
 
+The `persistTo` property defines where the local runtime is persisted to. This is only relevant when using `mode: local`. The default value is `.wrangler/state/v3` which means the files are safed to the directory `.wrangler/state/v3` relative to your astro dev execution path. This is the default path wrangler uses, so both tools should be able to pickup and work with the local state. You can change this to any other directory, however this directory should be included in the `.gitignore`.
```
sarah11918, line 202:
```suggestion
The `persistTo` property defines where the local runtime is persisted to when using `mode: local`. This value is a directory relative to your `astro dev` execution path.

The default value is set to `.wrangler/state/v3` to match the default path Wrangler uses. This means both tools are able to access and use the local state. 

If you change this value to any other directory, be sure to include it in `.gitignore`.
```
Took a shot from a straight language perspective, since I'm not familiar with the process itself. Make these sentences correct, but my thoughts are:

- that was a big, dense paragraph. Splitting up into what it is, the info about the default, and configuring something other than the default

- unless these are precise or common technical terms, "pick up" and "work with" are slightly colloquial English phrase that will be less easy to interpret or translate by non-native speakers. Precise terms, even if they sound drier or "more difficult", are often more accessible because they will be more standard and easier to look up if necessary. I inferred this meant something like "access and use" but choose whatever precisely is happening here. "detect"? "find"? whatever is correct!

- You would hate the way to properly use "however" in a sentence, so I rewrote to avoid.  :joy: To use properly it is written as:

> ... any other directory; however, this directory ...

So, adjust to taste, but I think by virtue of saying "if you do" implies that you CAN do it, so maybe you don't need to both say that it's possible, and then what you have to do if you choose to do it.

sarah11918 (commented):
Reviewed with some thoughts for your consideration, @alexanderniebuhr ! :partying_face:

packages/cloudflare/src/index.ts
```diff
@@ -59,6 +65,24 @@ interface BuildConfig {
...
+	}
+}
+
+const RUNTIME_WARNING = `You are using the deprecated string format for the \`runtime\` API. This will be phased out in the future. Please update to the new format for better support. 
+
+Example:
+Old Format: 'local'
+New Format: { mode: 'local' }
+
+Please refer to our updated documentation for more details on the new format.`;
```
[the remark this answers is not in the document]
alexanderniebuhr: Links can only be full URLs, so I added it to the end.

packages/cloudflare/README.md
alexanderniebuhr: Totally agree, the only thing I changed is, that the user even has to put the default directory to `.gitignore` themself.

packages/cloudflare/README.md
```diff
@@ -186,17 +186,25 @@ export default defineConfig({
...
+The `mode` property defines how the runtime is added to `astro dev`:
+
 - `local`: uses bindings mocking and locally static placeholders
 - `off`: no access to the Cloudflare runtime using `astro dev`. You can alternatively use [Preview with Wrangler](#preview-with-wrangler)
 
+The `persistTo` property defines where the local runtime is persisted to when using `mode: local`. This value is a directory relative to your `astro dev` execution path.
+
+The default value is set to `.wrangler/state/v3` to match the default path Wrangler uses. This means both tools are able to access and use the local state.
+
+The `.wrangler` directory or if you change this value to any other directory, be sure to include it in `.gitignore`.
```
sarah11918, line 206:
```suggestion
Whichever directory is set in `persistTo`, `.wrangler` or your custom value, included in `.gitignore`.
```
Is this true? Even if you use the default, you have to also update `.gitignore`? It doesn't handle the default one for you?

If so, then I'd suggest something closer to this.

alexanderniebuhr: Yes currently there is no way to automatically update `.gitignore` file on adapter install, so the user would have to handle that.

(https://discord.com/channels/830184174198718474/845430950191038464/1164249954836041848)

.changeset/rich-foxes-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Improves the developer experience, when using Cloudflare runtime with `astro dev`
```
sarah11918, line 5: t ```suggestion
Adds a new property `persistTo`, which allows .... when using Cloudflare runtime with `astro dev`
```
You didn't think I was going to let you get away with that, did you?  :smile:

Also, isn't there a breaking change in here because a value's format has changed from a string to an object?

Comment (alexanderniebuhr) [reply .changeset/rich-foxes-deliver.md:5]: I updated the Changeset, but would like for you to check it from the language aspect again. :)

File: .changeset/rich-foxes-deliver.md [hunk 12]
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Adds a new property `persistTo`, which allows to set the directory for local state files when using Cloudflare runtime with `astro dev`. This is useful when you want to persist state between restarts of the dev server, for example when using KV, D1, R2 to store data. Additionally, adds a warning, if the old configuration format is used. The current format is `runtime: { mode: 'off' | 'local', persistTo: string }`. See [runtime documentation](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime) for more information.
```
sarah11918, line 5:
```suggestion
Adds a new property `persistTo` which allows setting the directory for local state files when using Cloudflare runtime with `astro dev`. This is useful when you want to persist state between restarts of the dev server, for example when using KV, D1, R2 to store data. 

Additionally, updates the format of the `runtime` configuration and adds a warning when the deprecated format is used. The current format is  now `runtime: { mode: 'off' | 'local', persistTo: string }`. See [runtime documentation](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime) for more information.
```
Something like this, maybe? I'd at least split these into two paragraphs.

.changeset/rich-foxes-deliver.md
```diff
@@ -0,0 +1,7 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Adds a new property `persistTo` which allows setting the directory for local state files when using Cloudflare runtime with `astro dev`. This is useful when you want to persist state between restarts of the dev server, for example when using KV, D1, R2 to store data.
+
+Additionally, updates the format of the `runtime` configuration and adds a warning when the deprecated format is used. The current format is  now `runtime: { mode: 'off' | 'local', persistTo: string }`. See [runtime documentation](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime) for more information.
```
sarah11918, line 7:
```suggestion
Additionally, updates the format of the `runtime` configuration and adds a warning when the deprecated format is used. The current format is now `runtime: { mode: 'off' | 'local', persistTo: string }`. See [runtime documentation](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#runtime) for more information.
```
Stupid double space

alexanderniebuhr: Good catch!

PR #33, opened 2023-10-20
fix(cloudflare): astro asset image service

alexanderniebuhr:
## Changes

- this fixes a regression introduced 
- related to https://github.com/withastro/astro/pull/8880
- for details see messages after: https://discord.com/channels/830184174198718474/845430950191038464/1165031078722011238

### DIFF
--- a/.changeset/nice-impalas-whisper.md
+++ b/.changeset/nice-impalas-whisper.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fixes regression in Astro asset's image service
+Fixes override of a valid `astro:assets` image service configuration. Now overrides are only applied when the configuration is known to be incompatible with Cloudflare.
--- a/.changeset/nice-impalas-whisper.md
+++ b/.changeset/nice-impalas-whisper.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes regression in Astro asset's image service
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -1,6 +1,7 @@
 import type { AstroConfig, AstroIntegration, RouteData } from 'astro';
 
 import { createRedirectsFromAstroRoutes } from '@astrojs/underscore-redirects';
+import { passthroughImageService } from 'astro/config';
 import { AstroError } from 'astro/errors';
 import esbuild from 'esbuild';
 import { Miniflare } from 'miniflare';
@@ -104,7 +105,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 	return {
 		name: '@astrojs/cloudflare',
 		hooks: {
-			'astro:config:setup': ({ config, updateConfig }) => {
+			'astro:config:setup': ({ config, updateConfig, logger }) => {
 				let imageConfigOverwrite = false;
 				if (
 					config.image.service.entrypoint === 'astro/assets/services/sharp' ||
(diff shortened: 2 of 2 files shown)


.changeset/nice-impalas-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes regression in Astro asset's image service
```
sarah11918, line 5:
```suggestion
Fixes a regression in `astro:assets` where.... (explain thing that happened here)
```
If there was a regression, it should be possible to say what happened?

But, looking at the error message, it says that it has switched you to the noop image service? This seems like a bigger deal. So is what's really happening here more like:

> Identifies when image optimization is not possible and configures a `noop` image service for the user with a warning

?

dreyfus92, line 5:
```suggestion
Fixes a regression in `@astrojs/cloudflare` adapter where it would overwrite the image configuration even when compatible with Cloudflare Images. The adapter now only overwrites the configuration when known to be incompatible, and has moved the compatibility logic into the adapter itself.
```

dreyfus92 (changes requested):
I hope this helps @alexanderniebuhr

.changeset/nice-impalas-whisper.md
VoxelMC: Perhaps, if I may weigh in here, here's a couple I thought of:

```
1. Fixes accidental overwrite of a valid `astro:assets` image service configuration.
2. `astro:assets` image service now only overrides invalid entry point configurations.
```

Hopefully either of these helps!

packages/cloudflare/src/index.ts
```diff
@@ -104,7 +105,17 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 		hooks: {
-			'astro:config:setup': ({ config, updateConfig }) => {
+			'astro:config:setup': ({ config, updateConfig, logger }) => {
+				let imageConfigOverwrite = false;
+				if (
+					config.image.service.entrypoint === 'astro/assets/services/sharp' ||
+					config.image.service.entrypoint === 'astro/assets/services/squoosh'
+				) {
+					logger.warn(
+						`Does not support image optimization. To allow your project to build with the original, unoptimized images, the image service has been automatically switched to the 'noop' option. See https://docs.astro.build/en/reference/configuration-reference/#imageservice`
```
VoxelMC, line 115:
```suggestion
						`This entry point configuration does not support image optimization. To allow your project to build with the original, unoptimized images, the image service has been automatically switched to the 'noop' option. See https://docs.astro.build/en/reference/configuration-reference/#imageservice`
```

Or, whatever the correct terminology is, e.g. `image provider`.

sarah11918:
I like everyone's suggestions!  😄  Some include more detail than others, and I honestly think either way works. (I do have a preference for starting with a verb like "Fixes" though, as I think that sets the reader up for what kind of change this is, and how important it might be to their own code/project.)

@alexanderniebuhr pick what you like now that you've seen some suggestions that describe the problem that existed (that this PR addresses) and ping me again for a final proof!

dreyfus92 (approved):
LGTM ✅

sarah11918 (approved):
LGTM!

ematipico (approved):
It would be great if you could tell us how you tested the fix

alexanderniebuhr:
@ematipico Oops, my bad! I tested the changes locally using a demo project and sprinkled in some console.log statements for clarity. I'll definitely include those test descriptions in upcoming PRs. Thanks for the heads-up! 🚀

PR #34, opened 2023-10-20
feat(cloudflare): add astro assets image service

alexanderniebuhr:
## Changes

- refactor of automatic `noop` behavior
- added support for external paid Cloudflare Image Resizing Service
- Added a new option flag which will also help with https://github.com/withastro/adapters/pull/53
- package updates & test suite renames, **which shouldn't be in here** but which I needed to solve some very strange issue with mocha

## Testing

- added a new test based on the `astro:assets` test, however I still wish to test just the output of `updateConfig` if that is somehow possible

## Docs

- changesets
- readme

PR #53, opened 2023-11-05
fix(cloudflare): allow compile-time image optimization for prerendered pages

stancl:
## Changes

TLDR of the issue: the driver currently unnecessarily blocks image optimization even if it could be used just fine on static pages. The only issue is with including the image services in the SSR bundle.

I was more or less trying to reimplement the solution [discussed on Discord](https://discord.com/channels/830184174198718474/830184175176122389/1170827297452998738) but realized we don't actually need to re-export the services here, since the entire point of that was bypassing the if check in this code I believe.

So the only change now is adding an endpoint.

TODOs:
- [ ] Implement the endpoint so that it points to an existing file
- [ ] Not sure how this will work with SSR? We want to use the noop driver on SSR `<Image>` tags. Maybe the endpoint can handle that? Otherwise we could maybe specify the *entry*point conditionally, i.e. something like this (I think?)
    ```js
    service: import.meta.env.DEV ? config.image.service : passthroughImageService()
    ```

@alexanderniebuhr

### DIFF
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -112,7 +112,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 					config.image.service.entrypoint === 'astro/assets/services/squoosh'
 				) {
 					logger.warn(
-						`The current configuration does not support image optimization. To allow your project to build with the original, unoptimized images, the image service has been automatically switched to the 'noop' option. See https://docs.astro.build/en/reference/configuration-reference/#imageservice`
+						`The current configuration does not support image optimization using SSR on edge. To allow your project to build with the original, unoptimized images, the image service has been automatically switched to the 'noop' option. See https://docs.astro.build/en/reference/configuration-reference/#imageservice`
 					);
 					imageConfigOverwrite = true;
 				}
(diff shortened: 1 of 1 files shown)


alexanderniebuhr:
@stancl, my apologies for the delayed response; I was finalizing a component to share here. From our Discord conversation, we recognized that incorporating a solution for optimization at build/compile time, as well as for server-side rendering (SSR), adds a level of complexity that may exceed the scope of this current PR.

Nonetheless, we agreed that it's beneficial to provide users with the flexibility to choose between build/compile time optimization and runtime optimization. In line with this, and leveraging my work on an external image service, I've introduced an adapter configuration flag that can easily accommodate new options. I propose we extend this configuration to support build/compile time optimization as a selectable option.

https://github.com/withastro/adapters/pull/34/files#diff-be36604f10bbd9205e3930ca4cdb26388354a780fe7a8407735067cbff7694ecR164-R172

Regarding implementation, we have a few paths forward:

a) You could rebase your branch onto the feature branch and incorporate this addition.
b) Alternatively, you might prefer to wait until the feature branch is merged into the main branch and then introduce your changes.
c) As a third option, I can create a new branch to integrate your changes.

I would advocate for options a or b to ensure that you receive full credit for this contribution, though I am mindful of the extra effort this may entail. Please share your preference on how you wish to proceed.

alexanderniebuhr:
Closing in favor of https://github.com/withastro/adapters/issues/57
I'll implement it in another branch.

PR #61, opened 2023-11-11
fix(cloudflare): code pattern cloudflare local runtime

### LINKED ISSUE #60: Memory / Assign leak using Cloudflare local runtime
We do have a memory leak, because we do start a new instance of miniflare for each request. Will have to rewrite that logic to be more robust using a factory class maybe

alexanderniebuhr:
### Current (Technical & Docs) Review Tasks:

- final review

### Old (Technical) Review Tasks:

- Review new (moved/refactored) code: [packages/cloudflare/src/utils/local-runtime.ts](https://github.com/withastro/adapters/pull/61/files#diff-e0f33e55875adb3acef2fd01f20772250c505c98513365708f381ca027ceeb9a)

### Old (DOCS) Review Task:
- Feedback on Adapter Readme: [packages/cloudflare/README.md](https://github.com/withastro/adapters/pull/61/files#diff-be36604f10bbd9205e3930ca4cdb26388354a780fe7a8407735067cbff7694ec)
- Feedback on Adapter API Options: [packages/cloudflare/test/fixtures/dev-runtime/astro.config.ts](https://github.com/withastro/adapters/pull/61/files#diff-a75d1bf23cee01d339cf906350f6a41ed82eec126c6e7c101b7dc505696be3fa)

### Old (Technical) Review Task:

- Feedback on Design/Code Architecture: [`packages/cloudflare/src/utils/local-runtime.ts`](https://github.com/withastro/adapters/pull/61/files#diff-e0f33e55875adb3acef2fd01f20772250c505c98513365708f381ca027ceeb9a)
- Feedback on user facing config API: [`packages/cloudflare/test/fixtures/dev-runtime/astro.config.mjs`](https://github.com/withastro/adapters/pull/61/files#diff-4fddfca88db0352e6d0ee9146983715f7523b759f3a57fb5e13d8d3b043a99bb)

---

#### Summary
This pull request introduces significant changes to our Cloudflare Runtime API to enhance clarity and alignment with Cloudflare's practices. Our current implementation, which revolves around the use of `wrangler.toml` for defining bindings & env vars, has been identified as a source of confusion, especially for users deploying to Cloudflare Pages. This update aims to resolve these ambiguities and streamline user experience.

#### Background
Feedback from multiple Astro users and a technical writer at Cloudflare revealed that our current approach is misleading. The use of `wrangler.toml` is appropriate only for deployments to Cloudflare Workers, yet the majority of our users deploy to Cloudflare Pages where this configuration is not applicable. This discrepancy has led to confusion, as our documentation inadvertently suggests practices unsupported by Cloudflare.

#### Key Changes
1. **Removal of `wrangler.toml` Support for Cloudflare Pages**: We are discontinuing the use of `wrangler.toml` for deployments to Cloudflare Pages. This change will not impact Cloudflare Workers users but requires action from Cloudflare Pages users who utilize local runtime support.

2. **API Syntax Update**: Previously, we transitioned the `runtime` configuration from a single string value to a more complex object. While we maintained backward compatibility with the string syntax in the past, this update will formally remove that. Users will need to adopt the new object syntax for runtime configuration, to be able to continue using it.

#### Implications
- **For Cloudflare Pages Users**: Users will need to update their configuration, moving away from `wrangler.toml` and adopting the new syntax for runtime configuration.

#### Rationale
This change is driven by the need to eliminate confusion and align our practices with Cloudflare's supported methods. Although the previous approach did not break functionality, it led to a misalignment with Cloudflare's documentation and best practices.

#### Process Management Improvement
In response to a critical issue, this pull request also addresses a significant leak in our existing system. We identified that a new miniflare/workerd process was being spawned with every page request. This led to multiple concurrent processes, which not only failed to share state but also competed for I/O operations on the same files. This inefficiency hindered clean shutdowns and overall system performance. The update rectifies this by ensuring more efficient process management, thereby enhancing system reliability and resource utilization.

#### Refactoring with Factory Function and Class Hierarchy
Another key improvement in this update is the introduction of a more sophisticated class structure using a factory function. This architectural change ensures that a new local runtime object is only created if one does not exist, thereby optimizing resource usage and process management. The design includes two distinct classes that extend from a shared base class. This approach maximizes code reuse while allowing for the necessary distinctions in handling `wrangler.toml` for Workers and inline adapter configuration for Pages. This structure aligns with the best practices discussed with Cloudflare and is adaptable to future configuration changes.

These new classes are engineered to be more robust in handling errors. They are designed to provide more informative and helpful log messages, which will aid users in diagnosing and understanding any issues that arise. This enhancement is crucial for improving user experience, particularly in scenarios where troubleshooting is required.

closes #60


alexanderniebuhr:
* main
  * **PR #61** <a href="https://app.graphite.dev/github/pr/withastro/adapters/61?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/withastro/adapters/61?utm_source=stack-comment).

alexanderniebuhr:
cc @dario-piotrowicz this is most likely what we are going with, I think.
Reviews from the team outstanding & docs still need to be written, but I wanted to let you know as soon as possible, to give you time to weigh in from Cloudflare's perspective :)

dario-piotrowicz:
@alexanderniebuhr thanks so much for pinging me 🙂, the API does look good to me 😄👍

bluwy:
The general direction here makes sense to me, but it doesn't look like it's ready to be reviewed and merge yet? The changesets, todo comments, and some parts of the code seem to not be done yet.

alexanderniebuhr:
> it doesn't look like it's ready to be reviewed and merge yet?

Basically it is. Once we have a "good to go" on the API design from all sides, I'll just do the changesets, little code formatting & organization and it's ready to get merged. _(I don't really expect the API to change afterwards)_

I just don't want to invest all of that time, if the API needs to be changed.

packages/cloudflare/src/utils/local-runtime.ts
```diff
@@ -0,0 +1,239 @@
...
+	public async getCF() {
+		await this._miniflare.ready;
+		const CF_ENDPOINT = 'https://workers.cloudflare.com/cf.json';
+		if (!this.cfObject) {
+			this.cfObject = await fetch(CF_ENDPOINT).then((res) => res.json());
+		}
+		return this.cfObject;
+	}
+
+	public async clean() {
```
ematipico, line 153: Maybe we can align this API to the name from miniflare, hence `dispose`.

packages/cloudflare/test/fixtures/dev-runtime/astro.config.ts
```diff
@@ -5,6 +5,19 @@ export default defineConfig({
 	adapter: cloudflare({
 		runtime: {
 			mode: 'local',
+			type: 'pages',
+			bindings: {
+				bindings: {
+					"COOL": "ME"
+				},
```
ematipico, line 12: The double `bindings` feel weird. I am sure there's a reason for that.

ematipico (commented):
Overall the refactor looks good and understandable. I have some reservations regarding the `bindings.bindings` thing, which seems weird.

packages/cloudflare/test/fixtures/dev-runtime/astro.config.ts
alexanderniebuhr: Hey @dario-piotrowicz, could you share your thoughts with an explanation? Your insights are always the best we can get, to make sure we understand Cloudlare's side!

Also, what's your take on an alternative approach to avoid the `bindings.bindings` scenario? Looking forward to your ideas!

dario-piotrowicz: Sure, thanks for pinging me 🙂 

I totally agree that `bindings.bindings` is quite awkward, it is just that this closely follows [Miniflare's API](https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare#core) (which has an unfortunate/confusing name for this property), so having this `bindings.bindings` structure is correct although awkward.

As alternatives there are two that I can think of, which basically involve diverging from the Miniflare's format:
 - you could rename the `bindings` field inside `bindings` something like `textBindings` or `jsonBindings` (that's what I did a couple of weeks ago [in our Next.js adapter](https://github.com/cloudflare/next-on-pages/blob/c9167ea4bd11cb235e1d0545f25f187fd83771fb/internal-packages/next-dev/src/index.ts#L31), but I am pretty sure I am going to revert that and going with the next alternative soon👇). This would allow you to keep things almost in sync with the Miniflare API and diverge minimally from it.
 - you could restructure the whole `bindings` object to follow a structure like `bindings?: Record<string, Binding>` (that I just learned that it's getting introduced [in wrangler](https://github.com/cloudflare/workers-sdk/blob/eaa16db25103d26ef31499634a31b86caccdf7aa/packages/wrangler/src/api/startDevWorker/types.ts#L34)) (in that case these text bindings would practically be [`var` type bindings](https://github.com/cloudflare/workers-sdk/blob/eaa16db25103d26ef31499634a31b86caccdf7aa/packages/wrangler/src/api/startDevWorker/types.ts#L216C1-L216C1)). This to me feels like the better approach but it would require you to take the user options and mapping them to the current Miniflare API. Also I am not sure how stable these APIs are in wrangler as it is current work in progress I believe. (PS: I _think_ that a future Miniflare version could use a structure similar to this as well)
 
As I said I think that the second alternative is better (sorry for not mentioning this any earlier, I actually had something like this in mind but I was totally unaware that the wrangler team already defined it there! 😓).
 
Please let me know what you think 🙂

alexanderniebuhr: Thank you for the insights; they help a lot.

Personally, I think it would be best to go with the second approach. Even though it might not be 100% stable and Cloudflare could change it again, I believe it is the more robust API decision. This approach offers us the most flexibility while still keeping us in sync with Cloudflare's design.

What do you think, @ematipico? Would this approach align with your preferences?

ematipico: Sounds good to me too!

packages/cloudflare/README.md
```diff
@@ -195,24 +195,42 @@ export default defineConfig({
 
 ### `runtime`
 
-`runtime: { mode: "off" | "local", persistTo: string }`
+`runtime: { mode: 'off' } | { mode: 'local'; type: 'pages'; persistTo?: string; bindings?: Record<string, CF_BINDING>} | { mode: 'local'; type: 'workers'; persistTo?: string; };`
```
alexanderniebuhr, line 198: **[DOCS]**
This is not even the full type. See `CF_BINDING`.
I personally don't think we can fit any helpful type here, my suggestion is to give it a type name and link that to the GitHub code? WDYT?

packages/cloudflare/README.md
```diff
@@ -222,21 +240,108 @@ import cloudflare from '@astrojs/cloudflare';
...
++         type: "r2",
++         bucket_name: "R2",
++       },
+        // example of a Durable Object binding
++       "DO": {
++         type: "durable-object",
++         className: "DO",
++       },
++     },
++   },
```
alexanderniebuhr, line 275: **[DOCS]**
We got feedback that we should document how to configure bindings, therefore here I've gone with one example per supported binding. WDYT?

packages/cloudflare/README.md
```diff
@@ -222,21 +240,108 @@ import cloudflare from '@astrojs/cloudflare';
...
+
+# example for R2 Binding
+[[r2_buckets]]
+binding = 'R2'
+bucket_name = 'R2'
+
+# example for Durable Object Binding
+[[durable_objects.bindings]]
+name = "DO"
+class_name = "DO"
```
alexanderniebuhr, line 334: [DOCS]
Similar as above. Got feedback people want to know how to use `wrangler.toml`

packages/cloudflare/test/fixtures/dev-runtime/astro.config.ts
```diff
@@ -5,6 +5,49 @@ export default defineConfig({
...
+				"DO": {
+					type: "durable-object",
+					className: "DO",
+				},
+				"DO_PROD": {
+					type: "durable-object",
+					className: "DO_PROD",
+				}
+			}
```
alexanderniebuhr, line 50: **[DOCS]**
Feedback on those API options. We do have some boundaries, since we want to keep as close as possible to Cloudflare's version, but we do have some flexibility _(I think)_. WDYT?

packages/cloudflare/README.md
sarah11918: [DOCS]

OK, if this seems overwhelming to you, my first suggestion is (reserve judgement on the heresy): What if someone didn't care about **types** at this point, and just wanted to know which values they can set here?

First, optimize for **usage** -- someone needs to be able to configure this. It's totally fine to take an additional breath (section) to specifically describe the types of the various options.

packages/cloudflare/README.md
```diff
@@ -195,24 +195,42 @@ export default defineConfig({
...
+- `pages`: Deployed to [Cloudflare Pages](https://pages.cloudflare.com/)
+- `workers`: Deployed to [Cloudflare Workers](https://workers.cloudflare.com/)
 
-- `local`: uses bindings mocking and locally static placeholders
-- `off`: no access to the Cloudflare runtime using `astro dev`. You can alternatively use [Preview with Wrangler](#preview-with-wrangler)
+The `mode` property defines what the runtime supports in `astro dev`:
 
-The `persistTo` property defines where the local runtime is persisted to when using `mode: local`. This value is a directory relative to your `astro dev` execution path.
+- `off`: no access to the runtime using `astro dev`. You can alternatively use [Preview with Wrangler](#preview-with-wrangler)
+- `local`: uses a local runtime powered by miniflare and workerd. This supports Cloudflare's Bindings currently. For unsupported features, you can alterntively use `moed: 'off'` in combination with [Preview with Wrangler](#preview-with-wrangler)
```
sarah11918, line 213:
```suggestion
- `local`: uses a local runtime powered by miniflare and workerd, which supports Cloudflare's Bindings. 
```
So, some problems understanding when I can/can't use this option that would ideally be addressed:

Forget for a moment that I might not know which features are unsupported (but, that's also a worry). Better to state when to use each one, rather than refer to the exact other configuration you already mentioned.

I'm also not sure, since you just said that Wrangler is an *alternative* to `off`, why I might now use `off` *with* Wrangler Preview.

Also, "Choose when" or "Only use if" kinds of statements are generally more helpful than saying "you can use the other configuration". Forget for a moment that in neither case do I know when or why I should/might choose Wrangler (though, I don't). I might prefer to add under `off` something about "use this for unsupported features such as...."

packages/cloudflare/README.md
```diff
@@ -195,24 +195,42 @@ export default defineConfig({
...
 
-The default value is set to `.wrangler/state/v3` to match the default path Wrangler uses. This means both tools are able to access and use the local state.
+The `persistTo` property defines where the local bindings state is saved to when using `mode: local`, so you don't get a fresh binding on every restart of the dev server. This value is a directory relative to your `astro dev` execution path. By default it is set to `.wrangler/state/v3` to allow usage of `wrangler` cli commands, e.g. for migrations. You don't need to check this directory into your repository.
 
-Whichever directory is set in `persistTo`, `.wrangler` or your custom value, must be added to `.gitignore`.
+## Cloudflare runtime
+
+The Cloudflare runtime gives you access to environment variables and Cloudflare bindings, read more about it in Cloudflare's [Workers](https://developers.cloudflare.com/workers/configuration/bindings/) and [Pages](https://developers.cloudflare.com/pages/platform/functions/bindings/) docs. Depending on your deployment type (`pages` or `workers`), you need to configure the bindings differently.
```
sarah11918, line 219:
```suggestion
The Cloudflare runtime gives you access to environment variables and Cloudflare bindings. You can find more information in Cloudflare's [Workers](https://developers.cloudflare.com/workers/configuration/bindings/) and [Pages](https://developers.cloudflare.com/pages/platform/functions/bindings/) docs. Depending on your deployment type (`pages` or `workers`), you need to configure the bindings differently.
```

packages/cloudflare/README.md
```diff
@@ -195,24 +195,42 @@ export default defineConfig({
...
+- [Cloudflare Workers KV](https://developers.cloudflare.com/kv/)
+- [Cloudflare D1](https://developers.cloudflare.com/d1/)
+- [Cloudflare R2](https://developers.cloudflare.com/r2/)
+- [Cloudflare Durable Objects](https://developers.cloudflare.com/durable-objects/)
+
+### Config
+
+#### Cloudflare Pages
+
+Cloudflare Pages does not support a configuration file. You need to configure the bindings using Cloudflare's Dashboard, if you deploy your pages project to production. To be able to access bindings locally, you need to configure them using the adpater's `runtime` option.
```
sarah11918, line 233:
```suggestion
Cloudflare Pages does not support a configuration file. 

To deploy your pages project to production, you need to configure the bindings using Cloudflare's Dashboard. To be able to access bindings locally, you need to configure them using the adpater's `runtime` option.
```

packages/cloudflare/README.md
```diff
@@ -222,21 +240,108 @@ import cloudflare from '@astrojs/cloudflare';
...
++   runtime: {
++     mode: 'local',
++     type: 'pages',
++     bindings: {
+        // example of a KV binding
++       "URL": {
++         type: "var",
++         id: "https://example.com",
++       },
+        // example of a KV binding
```
sarah11918, line 252: [DOCS]

just checking, BOTH of these say KV binding here.

packages/cloudflare/README.md
sarah11918: [DOCS]

These look very helpful! My only concern is not showing valid code. If you are only ever going to configure ONE of these at a time, then this example is misleading. (Admittedly, I don't know this, so I'm just checking)

It's not the biggest deal, but I would caution about showing multiple configs in the same file if this isn't how it works without some kind of warning that you'll only actually use one of these at the same time. If it is possible to configure multiple, then great!

packages/cloudflare/README.md
```diff
@@ -222,21 +240,108 @@ import cloudflare from '@astrojs/cloudflare';
...
+If you also need to define `secrets` in addition to enviroment variables, you need to add a `.dev.vars` file to the root of the Astro project:
 
-Gives you access to [environment variables](https://developers.cloudflare.com/pages/platform/functions/bindings/#environment-variables), and [Cloudflare bindings](https://developers.cloudflare.com/pages/platform/functions/bindings).
+```
+# .dev.vars
+DB_PASSWORD=myPassword
+```
 
-Currently supported bindings:
+If you want to use `wrangler` for cli commands, e.g. D1 migrations, you would also need to add a `wrangler.toml` to the root of the Astro project with the correct content. Consult [Cloudflare's documentation](https://developers.cloudflare.com/) for further details.
```
sarah11918, line 287:
```suggestion
If you want to use `wrangler` for cli commands, e.g. D1 migrations, you also need to add a `wrangler.toml` to the root of the Astro project with the correct content. Consult [Cloudflare's documentation](https://developers.cloudflare.com/) for further details.
```
"would" is almost never needed. So as a gut check, always remove it and see whether the sentence still works.

packages/cloudflare/README.md
```diff
@@ -222,21 +240,108 @@ import cloudflare from '@astrojs/cloudflare';
...
+[[d1_databases]]
+binding = "D1"
+database_name = "D1"
+database_id = "D1"
+preview_database_id = "D1"
+```
+
+#### Cloudflare Workers
+
+Compared to [Cloudfare Pages](#cloudflare-pages), for Cloudflare Workers you need to configure the bindings using a `wrangler.toml` config file in the root directory of your Astro project. This file is also needed for production deployments.
```
sarah11918, line 304: NOTE: I HAVE TO WALK AWAY BUT WANT TO EDIT THIS ONE THING

sarah11918, line 304:
```suggestion
Unlike for [Cloudfare Pages](#cloudflare-pages), for Cloudflare Workers you need to configure the bindings using a `wrangler.toml` config file in the root directory of your Astro project. This file is also needed for production deployments.
```

I don't think "compared to" is what you want here, because this didn't immediately give me a comparison. I suggested something above that I think might be close to what you mean, but the issue is that use of "compared to".

Note that it's not even really necessary to relate it to Cloudflare pages at all, since you can imagine some people will turn up on this page, look at the sidebar and think, "Oh, I want Cloudflare Workers!" and just click to get right here, never having read anything about Cloudflare Pages!  :smile:  You could totally just start this sentence at "For Cloudflare Workers, you need to..." and, that would be fine!

packages/cloudflare/README.md
alexanderniebuhr: You can configure any of them in any combination and any amount. 
All of the following examples are valid. My only concern now is that it is not clear that you can configure it like this. What is you impression? @sarah11918 

- `var + var + var`
- `var + var + D1 + KV + KV`
- `var + D1 + var + KV + R2 + var`

packages/cloudflare/README.md
alexanderniebuhr: Will do as discussed in Discord:

https://discord.com/channels/830184174198718474/1075516604428857344/1177231072132608050

packages/cloudflare/src/index.ts
```diff
@@ -45,19 +42,24 @@ type Options = {
...
+				mode: 'local';
+				type: 'pages';
+				persistTo?: string;
+				bindings?: Extract<RUNTIME, { type: 'pages' }>['bindings'];
+		  }
+		| {
+				mode: 'local';
+				type: 'workers';
+				persistTo?: string;
+		  };
```
alexanderniebuhr, line 62: `'remote' | { mode: 'remote' }` were non-functional / undocumented, so it is not an issue to remove them in this major directly

packages/cloudflare/src/tmp-types.d.ts
alexanderniebuhr, line 1: This file is basically just us, copy&pasting un-exposed types from Cloudflare. They will expose them through their packages at some point, that's the time when we can get rid of this file.

ematipico (approved):
Egregious work! 🚀

Issue #64, opened 2023-11-14
fix edge-case in `_routes.json` generation


alexanderniebuhr:
- related to https://github.com/withastro/adapters/issues/67
- information about this issue on main repo

alexanderniebuhr:
https://discord.com/channels/595317990191398933/1174021841023664288

lilnasy:
Ran some previews to single out the configurations/defaults that make `index.astro` render for unrouted paths.

Shared in a PR comment here: https://github.com/withastro/adapters/pull/66#discussion_r1394960160

PR #66, opened 2023-11-15
fix(cloudflare): Fix `404.astro` not being used when `routes.strategy` is set to `exclude`

ToxiWoxi:
## Changes

Turns out it was an adapter issue after all! Fixes withastro/adapters#67

Currently, missing routes in the Cloudflare adapter are responded to with blank 404 responses:
```js
return new Response(null, {
	status: 404,
	statusText: 'Not found',
});
```
Astro has its own error handler! We can use that by returning `undefined` for our `routeData` in our render function. This pr does that.

## Testing

By using `pnpm link` in the [latest](https://github.com/ToxiWoxi/astro-404-issue/tree/latest) branch of my 404 issue repo and going to `http://localhost:8788/thispagedoesnotexist` in my browser, I can confirm that these changes result in the #renderError method being invoked which allows 404 pages being rendered properly whether they're prerendered or server-rendered.

I've also set up 3 previews:
One that uses a [server-rendered 404 page](https://astro-404-ssr.pages.dev), another that uses a [prerendered 404 page](https://astro-404-prerendered.pages.dev), and a third that shows the [current behavior](https://latest.astro-404-issue.pages.dev).

## Docs

I don't recommend any changes to the docs.  
This PR should not change the behavior of any users as it returns 404 behavior back to normal, matching other adapters.  

## Issues
The current #renderError method in the main repo fetches the 404 page when the 404 page is prerendered. This is seen in a warning in Wrangler. It does not effect website visitors, but it does result in 2 subrequests per page load, which you can see in the `Functions metrics` on the Cloudflare dashboard. To avoid this, you can use the "include" routing strategy or use SSR on the 404 page.  
I may open an issue for this in the main Astro repo later.

### DIFF
--- a/.changeset/twelve-grapes-dream.md
+++ b/.changeset/twelve-grapes-dream.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fix 404 responses in "exclude" routing strategy
+Fixes an issue where `404.astro` was not used when `routes.strategy` was set to `exclude`
--- a/.changeset/twelve-grapes-dream.md
+++ b/.changeset/twelve-grapes-dream.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fixes a regression which caused the adapter to falsely return an empty 404 response, caused by a change in `app.match()` behavior, which was introduced in https://github.com/withastro/astro/pull/7754. We forget to update the Cloudflare adapter and this implements the necessary fix.
+Fixes a regression which caused the adapter to falsely return an empty 404 response, caused by an upstream change https://github.com/withastro/astro/pull/7754.
--- a/.changeset/twelve-grapes-dream.md
+++ b/.changeset/twelve-grapes-dream.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Fix 404 responses in "exclude" routing strategy
--- a/.changeset/twelve-grapes-dream.md
+++ b/.changeset/twelve-grapes-dream.md
@@ -1,5 +1,5 @@
 ---
-'@astrojs/cloudflare': minor
+'@astrojs/cloudflare': patch
 ---
 
 Fix 404 responses in "exclude" routing strategy
--- a/.changeset/twelve-grapes-dream.md
+++ b/.changeset/twelve-grapes-dream.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fixes an issue where `404.astro` was not used when `routes.strategy` was set to `exclude`
+Fixes a regression which caused the adapter to falsely return an empty 404 response, caused by a change in `app.match()` behavior, which was introduced in https://github.com/withastro/astro/pull/7754. We forget to update the Cloudflare adapter and this implements the necessary fix.
--- a/packages/cloudflare/src/entrypoints/server.advanced.ts
+++ b/packages/cloudflare/src/entrypoints/server.advanced.ts
@@ -39,35 +39,33 @@ export function createExports(manifest: SSRManifest) {
 			return env.ASSETS.fetch(request);
 		}
 
-		// REF: If routeData is undefined, the render method invokes the #renderError method
-		let routeData = app.match(request, { matchNotFound: true }) || undefined;
+		let routeData = app.match(request, { matchNotFound: true });
+			Reflect.set(
+				request,
+				Symbol.for('astro.clientAddress'),
+				request.headers.get('cf-connecting-ip')
+			);
 
-		Reflect.set(
-			request,
-			Symbol.for('astro.clientAddress'),
-			request.headers.get('cf-connecting-ip')
-		);
-
-		const locals: AdvancedRuntime = {
-			runtime: {
-				waitUntil: (promise: Promise<any>) => {
-					context.waitUntil(promise);
+			const locals: AdvancedRuntime = {
+				runtime: {
+					waitUntil: (promise: Promise<any>) => {
+						context.waitUntil(promise);
+					},
+					env: env,
+					cf: request.cf,
+					caches: caches as unknown as CacheStorage,
 				},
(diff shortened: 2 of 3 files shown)


lilnasy:
Adding some indentation back so that the diff is easier to review.

lilnasy:
Ah! So the adapter is the one returning the empty 404 responses. It is confusing because that's what astro does when there isn't a `404.astro` as well.

.changeset/twelve-grapes-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
```
lilnasy, line 2:
```suggestion
'@astrojs/cloudflare': pathc
```
This is somewhat of a drastic change of behavior but still a patch.

.changeset/twelve-grapes-dream.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix 404 responses in "exclude" routing strategy
```
lilnasy, line 5:
```suggestion
Fixes an issue where `404.astro` was not used.
```
Routing strategy is not directly involved, is it?
It only hid the issue when the platform could serve the prerendered `404.html` instead.

lilnasy (approved):
LGTM, thanks for looking into it!

A couple of suggestion on the changeset but not blocking.

ToxiWoxi:
Yeah! I was playing around with Astro's render function in the main repo and noticed that nothing changed when I modified the renderError method. Sure enough, it never got called.

lilnasy:
That's incredible! The monorepo and the codebase can be hard to work with but you got to the bottom of it.

.changeset/twelve-grapes-dream.md
ToxiWoxi: The other issues still persist in the include strategy. This appears to be due to the routes.json issues, so this may solve that once those are fixed, but it would require further testing.

lilnasy: That can be done with this npm version: `@astrojs/cloudflare@experimental--404-fix`. I'll try it in a bit.

lilnasy: The branch named `latest` on your repo: https://latest.xyz-col.pages.dev/
The branch named `latest-prerendered` on your repo: https://latest-prerendered.xyz-col.pages.dev/

Both are running your fix.
Not sure about what to look for after changing the routing strategy.

ToxiWoxi: After looking into it further myself, the issues with rendering 404 pages in the include strategy aren't actually related to the routes, but how Cloudflare treats apps without a 404.html. 

Before the change that implemented include mode, (withastro/astro#7846), the default behavior was similar to what exclude mode is now.  
In include mode, all non-defined routes use Cloudflare's static method. In this method, if there is no 404.html, then it's assumed that you're running a SPA and renders `index.html` instead (with status code 200). If the index route is server rendered, it instead renders an empty 404 error.
In exclude, all routes, excluding static files, are rendered through Astro. This is what allows not-found routes to be caught by Astro and rendered appropriately.

alexanderniebuhr actually found this out yesterday when starting to figure out #64 and posted a discord thread that lead to that conclusion.

ToxiWoxi: So to conclude, this PR would not be able to change how 404 responses are resolved in include mode as missing routes are not sent through Astro.

alexanderniebuhr (changes requested):
@ToxiWoxi thanks for the PR! 🚀 
I think I understood the change, but I'd like to test it a little bit myself _(therefore going to block this)_, because I don't think this should be the right approach. But let me get back once I know more.

alexanderniebuhr (commented):
Generally this can be a good change, because if `Astro.render()` handles cases where `routeData` is `undefined` we don't need to have the custom logic.

**However**, I still want to wait for information about the fact that `routeData` started to return `undefined` instead of matching the `404`, which in fact seems to be a regression, because the option `matchNotFound` should match it.

.changeset/twelve-grapes-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes an issue where `404.astro` was not used when `routes.strategy` was set to `exclude`
```
alexanderniebuhr, line 5: We might need to update the `changeset` again, once we know why `routeData` behavior changed. Waiting for clarification on that.

ToxiWoxi:
In the 404/500 refactor in withastro/astro#7754, `matchNotFound` was removed.  
They moved 404 handling from `app.match` to `app.render`.  
To use the new error handler, `routeData` must be `undefined`.  
This is done on line 150 of `astro/packages/astro/src/core/app/index.ts` (which you can look at [here](https://github.com/withastro/astro/blob/1862fb44eb1008e17222130ff74a7a589aacecfd/packages/astro/src/core/app/index.ts#L150))

alexanderniebuhr:
Interesting! Thanks for pointing this out. Wondering why other adapters where updated but not the Cloudflare one during that refactor.
I personally still think that is not the best call, but if it was changed in Astro, we do have to change it here too. I'll propose another `changeset`, because I think we need to make it more clear. I'll also let Docs check it, so we are safe.

@ToxiWoxi thanks for the contribution, and my apology for asking many questions, it is important to get down to the root and find out why the behavior changed without making sure the Cloudflare adapter stays compatible. I'll make sure this get's merged soon.

ToxiWoxi:
> my apology for asking many questions

Completely understand! There's a lot of jumping around you have to do to figure it all out and the diffs that caused the changes are fairly large. Thanks for everything!

alexanderniebuhr (approved):
We've discussed this on Discord and we'll go ahead with this solution.
Just giving docs some time to check the changeset, I'll merge it tomorrow anyways.

sarah11918 (approved):
Really nice changeset @alexanderniebuhr ! :partying_face:

PR #31, opened 2023-10-19
feat(netlify): v2 functions

lilnasy:
## Changes
- In October, Netlify introduced a new API for functions, called [Functions 2.0](https://www.netlify.com/blog/introducing-netlify-functions-2-0/). The more familiar API is now referred to as [lambda-compatible](https://www.netlify.com/blog/introducing-netlify-functions-2-0/#lambda-compatibility-mode).
- The new API is standards based so would be much easier to maintain (The server entrypoint is 4 lines of code!).
- It streams responses by default, allowing users to improve page load performance.

## Testing
_Pending_

## Docs
_Pending_

PR #84, opened 2023-12-04
feat(netlify): Netlify Adapter v4

Skn0tt:
## Changes

This PR overhauls the Netlify adapter, aiming to make best use of Netlifys platform & streamlining config.
Read the changeset for a detailed overview, but here's the main changes:

- Middleware runs on all pages, not just SSR routes, independent of config
- Netlify context automatically available via locals, no need for `netlify-edge-middleware.ts` file
- Support for Netlify Image CDN
- Removal of On-Demand Builders, because we recommend frameworks to use standard caching headers instead
- Removal of unneeded config options

## Testing
  
Tests were brought up-to-date.

## Docs

README was updated.


.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,87 @@
...
+Remove the `netlify-edge-middleware.ts` or `netlify-edge-middleware.js` file.
+In your codebase, change all usage of locals injected through that file to use `Astro.locals.netlify.context` instead.
+
+### Image CDN
+
+v4 of this adapter integrates your Astro site with Netlify [Image CDN](https://docs.netlify.com/image-cdn/overview/).
+This allows transforming images on-the-fly without impacting build times.
+It's implemented using an [Astro Image Service](https://docs.astro.build/en/reference/image-service-reference/), and enabled by default.
+
+## On-Demand Builders are no longer supported
```
lilnasy, line 43: On-demand builders have a special place in the astro community.

The vast majority of our users are concerned with static sites. When their sites become large, our users start looking around for options to reduce work done during build-time but are intimidated by the world of complexity that SSR introduces.

With that context, when I introduce builders to them and they ask, "Is it really as simple as enabling a config?", I have been able to say yes without elaboration.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,189 @@
...
+
+	async function writeSSRFunction() {
+		await writeFile(
+			new URL('./ssr.mjs', ssrOutputDir()),
+			`
+				import ssrRoute from './entry.mjs';
+				export default ssrRoute;
+				export const config = { name: "Astro SSR", generator: "@astrojs/netlify@${packageVersion}", path: "/*", preferStatic: true };
+			`
+		);
```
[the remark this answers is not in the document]
Skn0tt: This is needed because of the exported config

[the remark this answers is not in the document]
lilnasy: If enough users are on a netlify cli version where default export detection works, we could prevent creating this additional file. The "exports" field of the adapter API will allow you to describe config from "ssr-function.ts".
https://docs.astro.build/en/reference/adapter-reference/#exports

.changeset/ten-cougars-pretend.md
lilnasy: It does not sound like an issue sitting here at all, but "how much do I need to learn" is a thought users always have in my experience, and tiny things like this are the ones to make them decide one way or another.

lilnasy: I am not sure.

Personally, I'd want on-demand builders to stay. I would have removed the lambda compatible API in my PR too if the new API had a "persistent" mode.

We could add things by default but that would just be hiding complexity. If the stance from Netlify really is to remove builders at some point, the users will just have to read up more on server stuff and that's fine too.

Just meant to point out that this is bit of a downgrade in one aspect.

Skn0tt: > If the stance from Netlify really is to remove builders at some point

I don't see that happening anytime soon - we have happy customers relying on builders, and we'll keep supporting that. But we consciously haven't added support for it to V2 Functions, because we see the cache headers as the new thing. But while they're great tools for framework implementors, I agree they're pretty complex things for most framework users.

What I think we should do us to add a `cache` option to the Adapter config. It would serve as a replacement for the `builders` option that currently exists.
If the user sets `cache: true`, the generated code will apply some sane `CDN-Cache-Control` header that behaves similar to builders. If the user wants to customize it per-page, like with `setBuildersTtl`, they can override it by setting Cache-Control headers from within the page itself.

lilnasy: That sounds perfect.

One thing I am curious about is how long far-future cache stays. As a demo for ODM, I often share this: https://github.com/lilnasy/astro-netlify-isr. The rendered date of the live demo is for July. It is easy to explain it in terms of "static but built later". But I'm guessing we would need to document that far-future cache may be evicted sooner?

Skn0tt: Correct, that cache is evicted once the TTL elapses, or the CDN drops the cache for some other reason. So it's not as obvious as with ODBs when it's regenerated. But developers won't really need to think about this, so I don't think that needs to be explained in detail in docs.

I'll get to work on the `cache` option 👍

Skn0tt:
Tests are passing, we've addressed the ODB removal. Looking good from my end! Can't wait for your reviews, but take it slow. Curious about those announcements :D

lilnasy:
There seems to be a failure.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+	}
+
+	async function writeMiddleware(entrypoint: URL) {
+		await mkdir(middlewareOutputDir(), { recursive: true });
+		await writeFile(
+			new URL('./entry.mjs', middlewareOutputDir()),
+			`
+						import { onRequest } from "${fileURLToPath(entrypoint)}";
```
lilnasy, line 96:
```suggestion
						import { onRequest } from "${fileURLToPath(entrypoint).replaceAll('\\', '/')}";
```

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+
+						export const config = {
+							name: "Astro Middleware",
+							generator: "@astrojs/netlify@${packageVersion}",
+							path: "/*", excludedPath: ["/_astro/*", "/.netlify/images/*"]
+						};
+						`
+		);
+
+		// taking over bundling, because Netlify bundling trips over NPM modules
```
lilnasy, line 122: This was the major talking point over removing edge adapters to begin with. It's not just Netlify, any build tool would trip over NPM packages. Sooner or later every user would find a package that can't automatically be transformed.

The fact that edge middleware was opt-in kept these issues few and far between. I'm noticing that this is being enabled by default, which opens us up to that maintenance burden again.

This will likely be blocking.

Skn0tt: Thanks for flagging this in advance. I'd really like to have this enabled by default, because from my perspective Middleware only really makes sense when it runs close to the user - so it needs to be an edge function.

I hear the maintenance argument loud and clear. There's two things that should reduce the burden from this change:

First, we've started building NPM support for Edge Functions. It's currently in Beta (see [this post](https://answers.netlify.com/t/support-for-npm-modules-in-edge-functions/101645)), and already covers a bunch of usecases. This is maintained by us, so any bugs in there are on us. If we can make use of that, errors during bundling would visibly be shown as part of Netlify's build step, not as part of the framework build. This should move a sizeable part of bug reports towards Netlify's support, instead of the Astro folks.

Secondly - there will always be issues that are opened in this repository, or in the Astro Discord, that will place a burden on the Astro Team. But we're happy to help! We've agreed to help maintain this adapter - part of that is responding to issues etc. If you tag us on issues regarding NPM modules in Edge Functions, we'll happily deal with them! Same is true for Discord - you might have seen me jumping into some questions in the #netlify channel. I still need to figure out how to set up email notifications for that, but feel free to tag me on questions there as well - i'll be happy to help.

Does that make you feel OK with enabling Edge Functions by default?

The reason I put ESBuild into here, is that there's a standing bug around namespaced imports (like `astro/middleware`). I'll take a look at fixing those, that should also allow us to remove the esbuild step here - so there's one less thing to maintain regarding NPM packages.

.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,84 @@
...
+## Middleware without configuration
+
+In v3, [Middleware](https://docs.astro.build/en/guides/middleware/) was deployed to the SSR Function by default.
+Enabling the `edgeMiddleware` option would change this behaviour to deploy it using Netlify Edge Functions instead, improving
+page speed and enabling the middleware to run on prerendered sites.
+
+In v4, this is the default setting.
+If your Astro site makes use of Middleware,
+it's deployed to an Edge Function where it runs super close to your users and is also applied to static sites.
+You can use this to implement customization or authorization logic!
```
ematipico, line 21: I am not sure we should enable this by default. Edge functions are subject to billing, especially in the free tier. If the edge functions happen to use all the bandwidth of the user, they will start paying without even noticing. What do you think?

packages/netlify/src/ssr-function.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	function createHandler(integrationConfig: { cacheOnDemandPages: boolean }) {
+		return async function handler(request: Request, context: Context) {
+			const routeData = app.match(request);
+			Reflect.set(request, clientAddressSymbol, context.ip);
+
+			let locals: Record<string, unknown> = {};
+
+			if (request.headers.has('x-astro-locals')) {
+				locals = JSON.parse(request.headers.get('x-astro-locals')!);
```
ematipico, line 24: This operation can cause a parse error that we should catch and log the error that caused the issue.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+			`
+		);
+	}
+
+	async function writeMiddleware(entrypoint: URL) {
+		await mkdir(middlewareOutputDir(), { recursive: true });
+		await writeFile(
+			new URL('./entry.mjs', middlewareOutputDir()),
+			`
+						import { onRequest } from "${fileURLToPath(entrypoint).replaceAll('\\', '/')}";
```
ematipico, line 96: nit: Can we reduce a bit the indentation of the template string to help readability, please?

ematipico, line 96: When dealing with Windows, we found that the best compatible way is to use `.toString()` instead of `replaceAll`. I think should be able to pick up the `file://` protocol. Also, does `entrypoint` contain the `C:\\` in this case?

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+
+						export default async (request, context) => {
+							const ctx = createContext({ 
+								request,
+								params: {}
+							});
+							ctx.locals = { netlify: { context } }
+							const next = () => {
+								const { netlify, ...otherLocals } = ctx.locals;
+								request.headers.set("x-astro-locals", trySerializeLocals(otherLocals));
```
ematipico, line 107: `trySerializeLocals` can throw a parse error that we should catch.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+
+	async function writeSSRFunction() {
+		await writeFile(
+			new URL('./ssr.mjs', ssrOutputDir()),
+			`
+				import createSSRHandler from './entry.mjs';
+				export default createSSRHandler(${JSON.stringify({
+					cacheOnDemandPages: Boolean(integrationConfig?.cacheOnDemandPages),
+				})});
+				export const config = { name: "Astro SSR", generator: "@astrojs/netlify@${packageVersion}", path: "/*", preferStatic: true };
```
ematipico, line 86: What does `preferStatic` mean in this context?

packages/netlify/src/index.ts
ematipico: My **main** concern is around **billing**. In the past, we enabled Vercel features that affected user's tiers without knowing it. We reverted those changes. Maintenance burden comes second place, from my point of view.

I want to make sure that users know what they're doing. While Netlify offers `1M /month` of Edge Functions to the users when they use the Free Tier, I don't know how to translate these values, and maybe even the user doesn't know. Build billing is measured in time, so it's straightforward for the users, although the measure of the Edge Functions is a bit different, and it might be difficult to translate/understand for rookies (I am a rookie).

ematipico (commented):
Hi @Skn0tt , thank you very much for this great contribution! The code is definitely better, that's something you should be proud of! 

About the PR, would it be possible to expand the description of the PR and explain a bit more about the decisions that you made? Feel free to be technical as much as you can. I noticed a lot of information is in the changelog, but not in the description of the PR.

.changeset/ten-cougars-pretend.md
Skn0tt: The free tier includes 1M edge function executions (see https://www.netlify.com/pricing/#core-pricing-table), which should be more than enough. Also, if there's no Middleware in the Astro project, then no Edge Function is generated. I think it's fine to have this as a default.

packages/netlify/src/ssr-function.ts
Skn0tt: This header is set by the upstream edge function, so we can trust its format. If there's a parser error happening, there will be an exception showing in the Edge Function logs - that should be enough for debugging.

packages/netlify/src/index.ts
Skn0tt: good point! not sure what happened here. will do.

Skn0tt: In my local testing, `.toString()` didn't work - it trips over the `file://` protocol. The current code seems to pass tests on windows - could you outline a scenario where this breaks?

packages/netlify/src/index.ts
Skn0tt: It tells the Netlify CDN to prefer serving static files on the path. If we didn't have it, the `/*` path would mean that static assets like prerendered pages or CSS files are never served, and all requests would go to this function. That's not what we want, so we set `preferStatic: true`. It means that if there's a static file on the request path, that is served instead of the function.

This is relatively new and super useful for framework integrations. We haven't documented this publically yet, but that will come.

.changeset/ten-cougars-pretend.md
Skn0tt: The Astro docs say the following:

> Middleware allows you to intercept requests and responses and inject behaviors dynamically every time a page or endpoint is about to be rendered.
> ....
> Middleware is available in both SSG and SSR Astro projects.

To me, this reads like Middleware should be executed not only for on-demand-rendered pages, but also for static pages. The only way of implementing this on Netlify is Edge Functions, as Functions don't run in front of static pages.

packages/netlify/src/index.ts
Skn0tt: done in https://github.com/withastro/adapters/pull/84/commits/c6daeb07e041d4253303cd1d624ee47db2e935c9

packages/netlify/src/index.ts
Skn0tt: What should we do about the error, if it occurs?

packages/netlify/src/index.ts
Skn0tt: I understand your concern around billing, but in the case of Edge Functions it should be fine. Let me put that number into perspective:

1M / month means 1 million requests that run an edge function. Since Astro middleware runs on all pages, but not on static assets, that's 1 Million page views essentially. That's enough to cover one page view every 3 seconds, which should be more than enough for sites on the free tier.

If a site gets super popular, every million requests on top of that free million is 2$ - so even if your site gets one visit per second (super popular site already!), that's 4$ per month in overages.

At these numbers, I don't think anybody on the free tier would run into overages. And our billing data shows the same, it's *extremely* rare that a free tier customer uses more than 1M edge function executions.

packages/netlify/src/index.ts
ematipico: I think a `console.error` should be enough. What do you think? Will the user be able to see the stacktrace from their dashboard?

Skn0tt:
> About the PR, would it be possible to expand the description of the PR and explain a bit more about the decisions that you made? Feel free to be technical as much as you can.

The rationale behind the big decisions should be in the changelog. Happy to expand on  that, what are the specific decisions you're interested in? It looks like we're already discussing some of the things in code.

packages/netlify/src/index.ts
Skn0tt: If we logged the error and then continued the request, it would be missing locals that might be important for authentication & authorization. So in the case of locals not being serializable, we definitely need to throw an exception and abort the request.
I'd say we keep it like it is - the user will be able to see the stacktrace from their dashboard already, and the error message thrown by `trySerializeLocals` should be enough to debug what's happening.

packages/netlify/src/index.ts
Skn0tt: > I'll take a look at fixing those, that should also allow us to remove the esbuild step here - so there's one less thing to maintain regarding NPM packages.

I'll have to walk back on this one. I gave it a go in https://github.com/netlify/edge-bundler/pull/547, but this is very hard to build within the architecture we currently have for NPM support. We'll have to rethink our approach there, or find a way to piggy-back on Deno's NPM support.

Let's stick with the ESBuild step we have in there for now, and replace it once we have something.

Skn0tt:
No worries! Resolved the conflicts, tests look like they're still passing.

Skn0tt:
Congrats on releasing Astro 4!

Looking through the [changelog](https://github.com/withastro/astro/releases/tag/astro%404.0.0), the only thing that might require changes in this adapter is https://github.com/withastro/astro/pull/9199. But since the old signature is supported both on Astro 3 and 4, it sounds like we should keep using that signature until the next major bump - does that make sense?

Skn0tt:
Okidoke, will do that 👍

Skn0tt:
What's the best way of expressing that v4 of this adapter only works with v4 of Astro? peer dependencies?

alexanderniebuhr:
Yes, I'm not sure about the change, we would have to see what others say, but IIRC it should be backwards compatible. At least we normally try to keep the adapters backwards compatible one version, so if `astro.render` is not, maybe holding off is actually better.

Skn0tt:
After talking to an Astro user on Discord, and talking a bit more about this internally, we've decided *not to remove ODBs* in this version. I'll start working on re-adding the builders flag to the adapter tomorrow.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+				}
+
+				setAdapter({
+					name: '@astrojs/netlify',
+					serverEntrypoint: '@astrojs/netlify/ssr-function.js',
+					exports: ['default'],
+					adapterFeatures: {
+						functionPerRoute: false,
+						edgeMiddleware: true,
+					},
```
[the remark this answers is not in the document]
Skn0tt: The removal of `functionPerRoute` is explained here: https://github.com/withastro/adapters/pull/84/files#diff-61372ae322de214c90be2cb1659463fa884f96d43179cc607c9fb1cfc6abdce1R67

I've outlined why I think Middleware always needs to run at the edge here: https://github.com/withastro/adapters/pull/84#discussion_r1415353909

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+					supportedAstroFeatures: {
+						hybridOutput: 'stable',
+						staticOutput: 'stable',
+						serverOutput: 'stable',
+						assets: {
+							supportKind: 'stable',
```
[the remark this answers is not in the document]
Skn0tt: Gonna change it to experimental 👍

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+							// still using Netlify Image CDN instead
+							isSharpCompatible: true,
+							isSquooshCompatible: true,
```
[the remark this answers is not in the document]
Skn0tt: The Image Service essentially generates some URLs that point towards Netlifys Image CDN, so I think it should work fine in endpoints & middleware, yes!

packages/netlify/src/index.ts
Skn0tt: From my perspective, this feature is already opt in. If you don't use middleware in your project, no middleware will be deployed to Netlify. It's only when you use middleware that an Edge Function is deployed.

*If* a user writes middleware, we'll have to deploy that somewhere. That's either Functions or Edge Functions, and looking at [our pricing page](https://www.netlify.com/pricing/#core-pricing-table) you can see that Edge Functions is the cheaper option of the two. Plus it runs on static pages, which (as per Astro's documentation) is required for Middleware.

You've probably noticed i'm a bit picky on this one, because I want the adapter no more complex than necessary. The more config options, the more questions will come up and the bigger the support burden.

Could you outline what opt-in mechanism you would like to see to alleviate this? Happy to make any changes necessary for that.

packages/netlify/src/index.ts
```diff
@@ -1,2 +1,220 @@
...
+	 * ---
+	 * <p>{Date.now()}</p>
+	 * ```
+	 */
+	cacheOnDemandPages?: boolean;
+}
+
+export default function netlifyIntegration(
+	integrationConfig?: NetlifyIntegrationConfig
+): AstroIntegration {
```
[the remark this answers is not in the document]
Skn0tt: Are you sure here? [`AstroAdapter`](https://github.com/withastro/astro/blob/ab0281aee419e58c6079ca393987fe1ff0541dd5/packages/astro/src/%40types/astro.ts#L2021) doesn't have a `hooks` field, and your Adapters documentation mentions that adapters are integrations: https://docs.astro.build/en/reference/adapter-reference/#building-an-adapter
To me, it looks more like `AstroAdapter` is the type for the parameters of `setAdapter`.

packages/netlify/src/index.ts
ematipico: Let's remove the billing aspect and the advantages of running code on the edge. I don't doubt the statements.

There are some technical aspects

An edge function runs with a different runtime (Deno), and if the user uses code that isn't Deno-compatible (mostly some dependencies), this will be come source of errors. A user should be able to opt-out of edge middleware if this happens. 

Plus 

> not all users need access to Netlify context
> not all users require having edge middleware

That I mentioned before.

> Could you outline what opt-in mechanism you would like to see to alleviate this? 

The previous configuration was enough. It's just a flag that says:

```js
netlify({
	edgeMiddleware: false
})
```

packages/netlify/src/index.ts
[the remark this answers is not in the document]
Skn0tt: > The adapterFeatures should says that the adapter supports it, right?

Could you elaborate on that? My understanding is that by saying `functionPerRoute: false` in `adapterFeatures`, Astro will generate only one big bundle, instead of one bundle for each route. That's what we want. Does `adapterFeatures` do something different than I think?

packages/netlify/src/index.ts
ematipico: Oh, never mind. I just noticed now that you used `fileURLToPath`. In some other instance (vercel adapter and the previous of netlify adapter maybe) I used `URL#toString()`. I'd say it should be fine :)

packages/netlify/src/ssr-function.ts
```diff
@@ -0,0 +1,54 @@
...
+				const hasCacheControl = [
+					'Cache-Control',
+					'CDN-Cache-Control',
+					'Netlify-CDN-Cache-Control',
+				].some((header) => response.headers.has(header));
+
+				if (!hasCacheControl) {
+					response.headers.append('CDN-Cache-Control', 'public, max-age=31536000, must-revalidate');
+				}
+			}
```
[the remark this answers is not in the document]
Skn0tt: done in https://github.com/withastro/adapters/pull/84/commits/b82e72432a9676c2dd8c99b2a76eabb049109abf

packages/netlify/src/index.ts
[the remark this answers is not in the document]
ematipico: Oh yeah you're right, my bad 😓

alexanderniebuhr (changes requested):
Checking in. I'm seeing we already got some technical review. Docs review will most likely be done nest week :)

packages/netlify/src/index.ts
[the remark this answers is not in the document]
matthewp: Never mind me, I was confusing `adapterFeatures` with `supportedAstroFeatures` 🤦‍♂️

Skn0tt:
Sounds good! Really like how much feedback i'm getting, this is probably the swiftest frameworks PR review i've gotten yet :D

packages/netlify/src/index.ts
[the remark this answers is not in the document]
ematipico: Yeah, in this particular case we are talking about the image service :) 

Flagging as experimental has the only effect of flagging a warning in the user's console, saying that "things could change". 

```shell
The feature is experimental and subject to change (used by @astrojs/netlify)
```

packages/netlify/src/index.ts
ematipico: I can see the confusion! 😅 The middleware is **available** in both modes, but how it's used/executed is different (for reasons!) based on the mode.

The middleware in SSG is **executed** during the page generation. Once the page is generated, the middleware isn't there anymore, because we have a static page. If a user wants to have a protected static page, Astro (middleware) isn't the way. 

Instead in SSR, the middleware is executed at each request, and that's when a user might decide to have a protected page with authorization.

Skn0tt: Got it! This is not obvious from the docs. I see it says “ 
every time a page or endpoint is about to be *rendered*”, which technically complies with the behaviour you’re outlining - but considering how “Middleware” works in other frameworks, it would be good to explicitly call this out.

Right now, the docs show an Authentication usecase (see [here](https://docs.astro.build/en/guides/middleware/#chaining-middleware)), which you’re saying does *not* work without Edge Middleware. This adds to the confusion.

I’ll work on reintroducing the `edgeMiddleware` option later today. Should it be opt-in or opt-out?

This brings me to another thing that we might need to build: If I’m understanding you right, any Response headers added in a Middleware won’t be present on static pages, because it writes the files to disk and the headers are lost. Is there some way for adapters to hook into that and get access to these custom headers? We could emit them headers into the `_headers` file, so the CDN sets them on static assets, similar to `_redirects`.

packages/netlify/src/index.ts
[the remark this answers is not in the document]
Skn0tt: Then it sounds like a good idea to keep this as experimental, at least until Netlify Image CDN is out of beta 👍

packages/netlify/src/index.ts
Skn0tt: I've readded `edgeMiddleware` as opt-in in https://github.com/withastro/adapters/pull/84/commits/bd7dbba6c23e815f23bd73a69b43bf0504d81e00. Chose to go opt-in because the Vercel adapter also does that, and it's good to behave the same by default across different adapters.

There's a lot of nuance around the behaviour for prerendered pages here. With `edgeMiddleware: false`, locals are applied to rendering, but you can't implement access control using it - with `edgeMiddleware: true`,  you can implement access control but locals don't work. This is confusing for all SSR adapters, and it could be useful to explain more about the tradeoffs here in the Astro docs.

packages/netlify/package.json
```diff
@@ -19,11 +19,8 @@
   "homepage": "https://docs.astro.build/en/guides/integrations-guide/netlify/",
   "exports": {
     ".": "./dist/index.js",
-    "./functions": "./dist/integration-functions.js",
-    "./static": "./dist/integration-static.js",
-    "./netlify-functions.js": "./dist/netlify-functions.js",
-    "./edge-functions": "./dist/integration-edge-functions.js",
-    "./netlify-edge-functions.js": "./dist/netlify-edge-functions.js",
```
ematipico, line 26: At Astro we don't delete stuff from one major to another, but we first deprecate things in a major, and delete it in the next major.

This means that we **can't** remove these entry points. 

Two other important things:
- I would argue that `/static` and `/serverless` are more descriptive than `/ssr-function`
- Why did we remove `/static`? Does this mean that we removed static support in the netlify adapter?

Skn0tt: > I would argue that /static and /serverless are more descriptive than /ssr-function

`ssr-function` is not meant for users, but only for passing something to the Astro Adapter: https://github.com/Skn0tt/adapters/blob/bd7dbba6c23e815f23bd73a69b43bf0504d81e00/packages/netlify/src/index.ts#L194
In that usecase, i'd say it's pretty descriptive.

> Why did we remove /static? Does this mean that we removed static support in the netlify adapter?

The adapter still supports static deployments, it's all inside the one main entrypoint. For static sites, we skip generating lambdas, but still generate redirects:

https://github.com/Skn0tt/adapters/blob/bd7dbba6c23e815f23bd73a69b43bf0504d81e00/packages/netlify/src/index.ts#L218-L229

> At Astro we don't delete stuff from one major to another, but we first deprecate things in a major, and delete it in the next major.

Thanks for pointing that out! In the current adapter, there's exports `/functions`, `/edge-functions` and `/static`. I can point `/static` and `/functions` towards the main entrypoint, and add a deprecation warning.
For `/edge-functions`, we'd probably have to add support for deploying to edge functions - but that's good to have, anyways.

Skn0tt: I've re-added `/static` and `/functions` in https://github.com/withastro/adapters/pull/84/commits/580820c78c4127339b833f1464936d26e0dbfc2a.

For `/edge-functions`, I did update the adapter to support it in https://github.com/withastro/adapters/pull/84/commits/3a03eb80f4513099278d2be9cbe6a7e234b08e18. Then realised that v3 had an `exports` field for it, but the implementation was removed in https://github.com/withastro/adapters/commit/19ea8b7a1668939c275e94b94ae159a3f8617c55. So I think it's fine to leave that out here, as well.

Skn0tt:
> Do you think it's a good idea to offer a mocked netlify context in dev mode?

That would be nice, I guess! We already mock the context in `netlify dev`, would be good for that to extend into Astro dev. How could that be implemented?

lilnasy:
You'll have to add a vite dev server (connect server) middleware, and set `request[Symbol.for("astro.locals")]`. Here's what it looked like in my PR: https://github.com/withastro/adapters/pull/31/commits/53775ba518b99ac217c2929e5f288fab9506927b.

Skn0tt:
Wow, that's pretty cool! I'll definitely implement that.

Skn0tt:
Besides that, are there any other changes you'd like me to make @lilnasy @alexanderniebuhr @ematipico? From what i'm seeing, all the big ones are addressed and the only missing thing is the docs review. What else do we need to work towards merging?

lilnasy:
I will be testing out the functionality later today.

As for the big picture, I would be more conformable if the default story is "node only" - for both SSR and middleware, with edge the opt-in.

alexanderniebuhr:
Just checking in, I don't have anything to add for now. But I'll make sure to get the docs review this week for you! :)

packages/netlify/src/index.ts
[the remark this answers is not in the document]
Skn0tt: I realized the log line doesn't include any mention which feature this is about, so it's hard for users to interpret this warning. Here's a PR to include that: https://github.com/withastro/astro/pull/9414

Skn0tt:
I've implemented the mocked `context` in https://github.com/withastro/adapters/pull/84/commits/4d4c24570a60180fdc4489cacddecb28e3286a80 based off your commit @lilnays. Tried giving you credit for that using `Co-Authored-By`, but somehow GitHub didn't like it and isn't showing your face on the commit :/

> As for the big picture, I would be more conformable if the default story was "node only" - for both SSR and middleware, with edge the opt-in.

Cool, will do! I thought the edge was already opt-in, and docs + comments said that - but I noticed the actual code was making it opt-out. Fixed that in https://github.com/withastro/adapters/pull/84/commits/0ac9466fe19ce99760d742d8085e3d79c5f2e187.

packages/netlify/README.md
```diff
@@ -49,81 +48,91 @@ If you prefer to install the adapter manually instead, complete the following tw
...
+This will deploy your Middleware as an Edge Function, and run it on all routes - including prerendered pages.
+This also means that locals specified in the Middleware won't be applied to any prerendered pages, because they've already been fully rendered at build-time.
+
+### Image CDN
+
+This adapter integrates your site with [Netlify Image CDN](https://docs.netlify.com/image-cdn/), transforming images on-the-fly without impacting build times.
+It's implemented using an [Astro Image Service](https://docs.astro.build/en/reference/image-service-reference/) under the hood.
+
+> **Note**
+> This adapter does not support `image.domains` and `image.remotePatterns` config properties in your Astro config - use the `remote_images` field in `netlify.toml` instead: [Netlify Image CDN - Remote Path](https://docs.netlify.com/image-cdn/overview/#remote-path).
```
sarah11918, line 133:
```suggestion
### Netlify Image CDN support

This adapter uses the [Netlify Image CDN](https://docs.netlify.com/image-cdn/) to transform images on-the-fly without impacting build times.
It's implemented using an [Astro Image Service](https://docs.astro.build/en/reference/image-service-reference/) under the hood.

> **Note**
> This adapter does not support the `image.domains` and `image.remotePatterns` config properties in your Astro config. To [specify remote paths for Netlify Image CDN](https://docs.netlify.com/image-cdn/overview/#remote-path), use the `remote_images` field in `netlify.toml`.
```

packages/netlify/README.md
```diff
@@ -146,117 +155,40 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
...
-```sh
-netlify deploy --build
-```
-
-The [Netlify Blog post on Astro](https://www.netlify.com/blog/how-to-deploy-astro/) and the [Netlify Documentation](https://docs.netlify.com/integrations/frameworks/astro/) provide more information on how to use this integration to deploy to Netlify.
-
-## Configuration
-
-To configure this adapter, pass an object to the `netlify()` function call in `astro.config.mjs` - there's only one possible configuration option:
+### Caching On-Demand-Rendered Pages
```
sarah11918, line 158:
```suggestion
### Caching Pages
```

packages/netlify/README.md
```diff
@@ -146,117 +155,40 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
...
 
-### dist
+On-demand-rendered pages that don't contain dynamic content can be cached to improve performance and lower resource usage.
```
sarah11918, line 160:
```suggestion
On-demand rendered pages without any dynamic content can be cached to improve performance and lower resource usage.
```

packages/netlify/README.md
```diff
@@ -146,117 +155,40 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
...
+The easiest way to do this is to enable the `cacheOnDemandPages` option in the adapter, which will cache all on-demand-rendered pages for up to one year:
```
sarah11918, line 161:
```suggestion
Enabling the `cacheOnDemandPages` option in the adapter will cache all server-rendered pages for up to one year:
```

packages/netlify/README.md
```diff
@@ -146,117 +155,40 @@ Once you run `astro build` there will be a `dist/_redirects` file. Netlify will
...
   output: 'server',
   adapter: netlify({
-    dist: new URL('./dist/', import.meta.url),
+    cacheOnDemandPages: true
   }),
 });
 ```
 
-And then point to the dist in your `netlify.toml`:
+For advanced usage, you can change this on a per-page basis by adding caching headers to your response from within your pages:
```
sarah11918, line 173:
```suggestion
This can be changed on a per-page basis by adding caching headers to your response:
```

.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,69 @@
...
+### Image CDN
+
+v4 of this adapter integrates your Astro site with Netlify [Image CDN](https://docs.netlify.com/image-cdn/overview/).
+This allows transforming images on-the-fly without impacting build times.
+It's implemented using an [Astro Image Service](https://docs.astro.build/en/reference/image-service-reference/), and enabled by default.
+
+## On-Demand Builders are no longer supported
+
+On-Demand Builders (ODB) allows SSR-Rendered pages to be cached using a Time to Live (TTL) strategy.
+ODB continues to be supported by Netlify, but has since been replaced by the much more powerful
```
sarah11918, line 31:
```suggestion
ODB continues to be supported by Netlify
```
this contradicts the heading. If you want people using the replacement instead, just say that

.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,69 @@
...
+[Fine-Grained Cache Control](https://www.netlify.com/blog/swr-and-fine-grained-cache-control).
+
+In v3, you could deploy your SSR-Rendered Astro pages to ODBs by enabling the `builders` config option,
+and then specifying the TTL on a per-page basis.
+In v4, there's the `cacheOnDemandPages` option that works very similarly - take a look at the README to learn more.
```
sarah11918, line 36:
```suggestion
In v3, you could deploy your SSR-Rendered Astro pages to ODBs by enabling the `builders` config option,
and then specifying the TTL on a per-page basis.
In v4, a new `cacheOnDemandPages` option replaces this config option.
```

.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,69 @@
...
+-   builders: true
++   cacheOnDemandPages: true
+  }),
+});
+```
+
+## `functionPerRoute` was removed
+
+In v3, the `functionPerRoute` option allowed the SSR routes to be split up into multiple Netlify Functions.
+In theory, this reduced bundle sizes of each individual function, and the speed-up in code parsing should speed up cold starts.
```
sarah11918, line 55:
```suggestion
This reduced the bundle sizes of each individual function, with the intention of speeding up code parsing, and therefore the time of cold starts.
```

packages/netlify/README.md
```diff
@@ -49,81 +48,91 @@ If you prefer to install the adapter manually instead, complete the following tw
...
    ```
 
-### Run middleware in Edge Functions
+## Usage
 
-When deploying to Netlify Functions, you can choose to use an Edge Function to run your Astro middleware.
+[Read the full deployment guide here.](https://docs.astro.build/en/guides/deploy/netlify/)
 
-To enable this, set the `edgeMiddleware` config option to `true`:
+After [performing a build](https://docs.astro.build/en/guides/deploy/#building-your-site-locally) the `.netlify/` folder will contain [Netlify Functions](https://docs.netlify.com/functions/overview/) in the `.netlify/functions-internal/` and [Netlify Edge Functions](https://docs.netlify.com/edge-functions/overview/) in the `.netlify/edge-functions/` folder.
```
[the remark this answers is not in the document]
Skn0tt:
```suggestion
Follow the instructions to [build your site locally](https://docs.astro.build/en/guides/deploy/#building-your-site-locally). After building, you will have a `.netlify/` folder containing both [Netlify Functions](https://docs.netlify.com/functions/overview/) in the `.netlify/functions-internal/` folder and [Netlify Edge Functions](https://docs.netlify.com/edge-functions/overview/) in the`.netlify/edge-functions/` folder.
```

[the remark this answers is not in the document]
Skn0tt: There was one bit of prose enclosed in a code block in your suggestion (`the`), don't think that was on purpose. Fixed that in my suggestion.

packages/netlify/README.md
```diff
@@ -49,81 +48,91 @@ If you prefer to install the adapter manually instead, complete the following tw
...
-### Per-page functions
+declare namespace App {
+  interface Locals extends NetlifyLocals {
+    ...
+  }
+}
+```
 
-The Netlify adapter builds to a single function by default. Astro 2.7 added support for splitting your build into separate entry points per page. If you use this configuration, the Netlify adapter will generate a separate function for each page. This can help reduce the size of each function so they are only bundling code used on that page.
+It's only available during on-demand rendering.
```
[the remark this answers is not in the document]
Skn0tt: It's mainly for projects that use on-demand rendering, but they might have pages that are prerendered. I'm trying to say this is not available for prerendered pages, maybe this works better?

```suggestion
This is not available on prerendered pages.
```

packages/netlify/README.md
```diff
@@ -49,81 +48,91 @@ If you prefer to install the adapter manually instead, complete the following tw
...
 
-```js
+### Run Middleware in Edge Functions
+
+If you use Astro Middleware, the default behaviour is that it's applied to prerendered pages at build-time, and to on-demand-rendered pages at runtime.
+For prerendered pages, it can't be used to implement redirects, access control or custom response headers.
+If you need any of this, you should probably deploy run your Middleware on Netlify Edge Functions by enabling the `edgeMiddleware` option:
```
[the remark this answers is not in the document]
Skn0tt: Reads much better!

packages/netlify/README.md
```diff
@@ -49,81 +48,91 @@ If you prefer to install the adapter manually instead, complete the following tw
...
   adapter: netlify({
-    functionPerRoute: true,
++   edgeMiddleware: true,
   }),
 });
 ```
 
-### Static sites
+This will deploy your Middleware as an Edge Function, and run it on all routes - including prerendered pages.
+This also means that locals specified in the Middleware won't be applied to any prerendered pages, because they've already been fully rendered at build-time.
```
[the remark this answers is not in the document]
Skn0tt: available is correct! reads much better now :)

.changeset/ten-cougars-pretend.md
Skn0tt: Fair point! I've made some changes to this in https://github.com/withastro/adapters/pull/84/commits/b63cfcb76a28af009c90de5bebd3a7e274753d23 and 
 https://github.com/withastro/adapters/pull/84/commits/13a765983f4abcf09906e7d0ec477998ceef29a6 that should capture more accurately what's happening. How does that read to you?

Skn0tt:
@sarah11918 wow, what a review! I always love docs review because the prose looks so much better after that :D I've made most of the changes you requested, and left comments to the two remaining ones.

ematipico (approved):
Thank you for addressing all my comments!

alexanderniebuhr (approved):
Unblocking, since we have docs review

Skn0tt:
Awesome, thanks for approving! I think now we're only waiting for the docs copy to be finalised.

packages/netlify/README.md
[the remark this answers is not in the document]
sarah11918: Yes, this works!

.changeset/ten-cougars-pretend.md
```diff
@@ -0,0 +1,69 @@
...
+
+v4 of this adapter integrates your Astro site with Netlify [Image CDN](https://docs.netlify.com/image-cdn/overview/).
+This allows transforming images on-the-fly without impacting build times.
+It's implemented using an [Astro Image Service](https://docs.astro.build/en/reference/image-service-reference/), and enabled by default.
+
+## On-Demand Builders are no longer supported
+
+On-Demand Builders (ODB) allows SSR-Rendered pages to be cached using a Time to Live (TTL) strategy.
+While the Netlify platform continues to support existing pages with ODBs, there's now the much more powerful
+[Fine-Grained Cache Control](https://www.netlify.com/blog/swr-and-fine-grained-cache-control) that's recommended going forward.
```
sarah11918, line 32:
```suggestion
## Replacement for On-Demand Builders

On-Demand Builders (ODB) allows SSR-Rendered pages to be cached using a Time to Live (TTL) strategy.
While the Netlify platform continues to support existing pages with ODBs, we now recommend using the much more powerful
[Fine-Grained Cache Control](https://www.netlify.com/blog/swr-and-fine-grained-cache-control) going forward.
```
I still think the text contradicted the header, so I'd suggest something closer to this. But this is much better! If you have a specific link for updating/converting, I'd also suggest adding a line like:

> Read [Netlify's documentation for converting your on-demand builders]() to upgrade to new the recommended caching method.

Skn0tt: That's good, thank you! We don't have specific docs for that, sadly, so there's nothing to link to :/

packages/netlify/README.md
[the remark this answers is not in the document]
sarah11918: @Skn0tt - Just in case this comment got lost, I think it was my last outstanding comment! Otherwise, docs look good!

alexanderniebuhr:
@Skn0tt Just checking in, if you are ready, I'll do another quick sanity check later tonight and get this merged 🚀

Skn0tt:
Sounds great! Maybe we can merge today + release early monday, so we can respond better to questions in Discord.

alexanderniebuhr:
@Skn0tt No problem, I'll coordinate everything, so we can get the release on Monday morning CET timezone! :)

packages/netlify/src/index.ts
alexanderniebuhr: Just to check in here, if you still think Astro docs are confusing here, I would suggest open a Issue directly in the docs repo or discuss this in the docs channel on Discrod. I'm afraid, it'll get lost here!

alexanderniebuhr (approved):
I just did another run through the changes, and they lgtm.
Thank you for the PR! @Skn0tt 

I'll get this merged now and released on Monday morning! 🚀

Skn0tt:
Awesome, thanks for all the reviews! I'll keep an eye on Discord and GitHub for any issues that pop up. If I miss any, feel free to tag me in related issues - happy to help.

Issue #104, opened 2023-12-18
[netlify] TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file://.../node_modules/@astrojs/netlify/package.json" needs an import assertion of type "json"


Skn0tt:
### Astro Info

```block
not needed
```

### Describe the Bug

See https://github.com/withastro/adapters/issues/3#issuecomment-1860791565

We're importing a JSON file without a type assertion. Some versions of Node.js don't like that.

### What's the expected result?

Importing the adapter should be fine.

### Link to Minimal Reproducible Example

not needed

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #105, opened 2023-12-18
fix: use fs read instead of JSON import

Skn0tt:
## Changes

Closes https://github.com/withastro/adapters/issues/104.

We were crashing when importing the adapter because importing JSON isn't supported on all versions of Node.js. This PR replaces it with `readFile` instead. We could've used import assertions instead, but then we'd also need to change some typescript settings which is risky.

## Testing

tested locally.

## Docs

no docs needed, it's a fix.

### DIFF
--- a/.changeset/fluffy-spoons-cheat.md
+++ b/.changeset/fluffy-spoons-cheat.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+fix crash when reading package.json version field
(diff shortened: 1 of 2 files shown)


alexanderniebuhr:
@Skn0tt Thanks for the fix. I'm cutting a patch release now. Just a general information, we tend to keep the format of the changeset in the present tense _(sometimes we don't catch that in reviews, and it is not a hard rule)_, but just so you know:

```
---
'@astrojs/netlify': patch
---

Fixes a regression introduced by v4, which caused a crash in some node versions
```

Skn0tt:
Thanks for letting me know, will try to remember it next time :D

PR #111, opened 2023-12-20
chore: use Biome as linter

ematipico:
## Changes

This PR uses Biome as linter. I applied safe and unsafe fixes. I suppressed some rules where we don't have enough information or where things were already suppressed with eslint.

## Testing

Current CI should work

## Docs

N/A


biome.json
```diff
@@ -29,6 +29,12 @@
     "enabled": true
   },
   "linter": {
-    "enabled": false
+    "enabled": true,
+    "rules": {
+      "suspicious": {
+        "noConsoleLog": "off",
```
alexanderniebuhr, line 35: Can we still get a `warn`?

biome.json
```diff
@@ -29,6 +29,12 @@
...
+        "noExplicitAny": "off"
```
alexanderniebuhr, line 36: What is the reason to disable this rule?

package.json
```diff
@@ -12,10 +12,11 @@
     "build": "turbo run build --filter=\"@astrojs/*\"",
     "build:ci": "turbo run build:ci --filter=\"@astrojs/*\"",
     "format:ci": "pnpm run format",
-    "format": "biome check --apply ./ && prettier -w \"**/*\" --ignore-unknown --cache",
+    "format": "biome check --linter-enabled=false --apply ./ && prettier -w \"**/*\" --ignore-unknown --cache",
```
alexanderniebuhr, line 15: Pls help me understand `biome`, isn't there a cli command which only formats, `biome format`?

packages/netlify/src/ssr-function.ts
```diff
@@ -21,6 +21,7 @@ export const createExports = (manifest: SSRManifest, _args: Args) => {
 			let locals: Record<string, unknown> = {};
 
 			if (request.headers.has('x-astro-locals')) {
+				// biome-ignore lint/style/noNonNullAssertion: safe because we checked before
```
alexanderniebuhr, line 24: I would love if typescript would be able to handle those checks on Maps....

alexanderniebuhr (approved):
@ematipico Thanks for taking the time and helping with this!
I'm in favor of this and thing it is great. 🚀 

All the changes lgtm!

**However,** I would like to see that we replace `eslint` completely. Can you give some more reasons, why we still need `eslint`?

packages/netlify/src/ssr-function.ts
ematipico: I can change the code. The `.get` method returns undefined, so we can check the result. We wouldn't need the bang operator

biome.json
ematipico: You would get a bunch of warnings, which are annoying I felt annoying. Ideally, we could use an override, but they are currently buggy. Are you still happy with a warning?

biome.json
ematipico: There are a bunch of `any` that I didn't want to touch, so I turned off the rule. How do you want to proceed?

package.json
ematipico: Yeah it does, but it **doesn't** sort the imports. To avail of import sorting, you need the `check` command. With this change, format code and sort the imports. The linter is disabled.

biome.json
alexanderniebuhr: Oh I didn't know we had them :/
Let's keep the rule `off` for now, and I'll double check the code, to get rid of those `any`s

ematipico:
> Can you give some more reasons, why we still need eslint?

Unfortunately, Biome linter isn't on par with **@typescript-eslint** because it cannot retrieve type information. Hence, I advise keeping it, but only the rules that require type information.  For the rest, I think you can rely on Biome.

packages/netlify/src/ssr-function.ts
ematipico: Done in [`0b9fc0e` (#111)](https://github.com/withastro/adapters/pull/111/commits/0b9fc0e2b297dddd50e613519061b530d86cbead)

biome.json
alexanderniebuhr: Hm that is a tricky one. In general we don't really want to use `console.log()` for anything user-facing, everything user-facing should probably use astro's logger or error. So getting a warning, sounds feasible. However I was also annoyed by it before, so disabling this makes also sense.

I'm happy both ways, so maybe keep it off for now!

packages/netlify/src/ssr-function.ts
alexanderniebuhr: Oh yeah that would work, I think rewriting the code makes sense.

I just think Typescript should support Map's `.has()`

alexanderniebuhr:
> [...] Biome linter isn't on par with typescript-eslint [...], I advise keeping [eslint], but only the rules that require type information. For the rest, I think you can rely on Biome.

Alright, sounds good for me. I really would love to get rid of eslint, I wonder how valuable those rules with type information are, but nothing to worry about in this PR!

ematipico:
> I wonder how valuable those rules with type information are

One that I deem really important is: https://typescript-eslint.io/rules/no-floating-promises/

alexanderniebuhr (approved):

Issue #118, opened 2023-12-29
Netlify adapter, can't opt out image service 


ematipico:
### Astro Info

```block
main
```

### Describe the Bug

The new version of Netlify adapter doesn't allow to opt out of the image service used by Netlify.

### What's the expected result?

Like the Vercel adapter, I would expect an option, because the Netlify image service doesn't support remote patterns **out of the box**.

### Link to Minimal Reproducible Example

None

### Participation

- [ ] I am willing to submit a pull request for this issue.

alexanderniebuhr:
I agree, having a option to opt-in / opt-out would be preferable.

cc @Skn0tt

Skn0tt:
Agreed! I'll work on this.

Skn0tt:
Here's the PR: https://github.com/withastro/adapters/pull/120

PR #120, opened 2024-01-02
[netlify] allow opting out of Image CDN

Skn0tt:
## Changes

As per https://github.com/withastro/adapters/issues/118, this adds an opt-out option for Image CDN.

## Testing

Added new unit tests.

## Docs

This is a new config option. We should probably mention it in the Astro integration docs.


packages/netlify/src/index.ts
```diff
@@ -52,6 +52,16 @@ export interface NetlifyIntegrationConfig {
...
 	 */
 	edgeMiddleware?: boolean;
+
+	/**
+	 * If enabled, Netlify Image CDN is used for image optimization.
+	 * This transforms images on-the-fly without impacting build times.
+	 * 
+	 * If disabled, Astro's built-in image optimization is run at build-time instead.
+	 * 
+	 * @default enabled
```
ematipico, line 62:
```suggestion
	 * @default {true}
```

packages/netlify/src/index.ts
```diff
@@ -242,7 +254,7 @@ export default function netlifyIntegration(
 					},
 					image: {
 						service: {
-							entrypoint: isRunningInNetlify ? '@astrojs/netlify/image-service.js' : undefined,
+							entrypoint: enableImageCDN ? '@astrojs/netlify/image-service.js' : undefined,
```
ematipico, line 257: Is it safe to set the service to `undefined`? Shouldn't we set the service only if `enableImageCDN` is `true`?

packages/netlify/test/functions/fixtures/middleware/astro.config.mjs
```diff
@@ -5,6 +5,7 @@ export default defineConfig({
   output: 'server',
   adapter: netlify({
     edgeMiddleware: process.env.EDGE_MIDDLEWARE === 'true',
+    imageCDN: process.env.DISABLE_IMAGE_CDN ? false : undefined,
```
ematipico, line 8: Should this be `=== "true"`?

Skn0tt: No, because then it would be `true` (aka: explicitly enabled) in cases where the env var isn't set. I want to test the default case `undefined`, where nothing was specified.

packages/netlify/src/index.ts
Skn0tt: We're also using enabled/disabled wording for the comment above. Personally, I prefer the english wording here as opposed to code syntax - but it's a preference thing. Is there a guideline on this in the Astro docs? cc @sarah11918

packages/netlify/src/index.ts
alexanderniebuhr: IIFRC, `undefined` equals the default one

packages/netlify/src/index.ts
ematipico: This is JSDoc, which wants a **value**: https://jsdoc.app/tags-default

In this case, `enabled` isn't a valid value for this property

ematipico (approved):
Thank you for addressing this so quickly!

packages/netlify/src/index.ts
Skn0tt: Fair point - gonna change this here and in the other comment.

packages/netlify/src/index.ts
```diff
@@ -49,9 +49,19 @@ export interface NetlifyIntegrationConfig {
...
+	 * @default {false}
 	 */
 	edgeMiddleware?: boolean;
+
+	/**
+	 * If enabled, Netlify Image CDN is used for image optimization.
+	 * This transforms images on-the-fly without impacting build times.
+	 * 
+	 * If disabled, Astro's built-in image optimization is run at build-time instead.
+	 *
```
alexanderniebuhr, line 61: I don't think this description is 100% correct, but it is also not blocking.

PR #127, opened 2024-01-09
fix(netlify): only cache GET/HEAD methods

Skn0tt:
## Changes

If `cacheOnDemandPages` is enabled, we're currently adding caching headers to all responses, including POST responses. Caching POST requests is unconventional, and shouldn't be done by default!

This PR makes it so the adapter only adds caching headers for GET and HEAD requests.

## Testing

No test added because there's no test for `cacheOnDemandPages` yet - let me know if I should add one.

## Docs

Don't think this needs docs.

### DIFF
--- a/.changeset/dry-rules-bathe.md
+++ b/.changeset/dry-rules-bathe.md
@@ -2,4 +2,4 @@
 '@astrojs/netlify': patch
 ---
 
-Fixes cacheOnDemandPages setting to only cache GET/HEAD requests by default
+Updates the behavior of the `cacheOnDemandPages` setting to only cache GET/HEAD requests by default
--- a/.changeset/dry-rules-bathe.md
+++ b/.changeset/dry-rules-bathe.md
@@ -2,4 +2,4 @@
 '@astrojs/netlify': patch
 ---
 
-only cache GET/HEAD requests with cacheOnDemandPages
+Fixes cacheOnDemandPages setting to only cache GET/HEAD requests by default
--- a/.changeset/dry-rules-bathe.md
+++ b/.changeset/dry-rules-bathe.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+only cache GET/HEAD requests with cacheOnDemandPages
--- a/packages/netlify/src/ssr-function.ts
+++ b/packages/netlify/src/ssr-function.ts
@@ -36,6 +36,8 @@ export const createExports = (manifest: SSRManifest, _args: Args) => {
 			}
 
 			if (integrationConfig.cacheOnDemandPages) {
+				const isCacheableMethod = ['GET', 'HEAD'].includes(request.method);
+
 				// any user-provided Cache-Control headers take precedence
 				const hasCacheControl = [
 					'Cache-Control',
@@ -43,7 +45,7 @@ export const createExports = (manifest: SSRManifest, _args: Args) => {
 					'Netlify-CDN-Cache-Control',
 				].some((header) => response.headers.has(header));
 
-				if (!hasCacheControl) {
+				if (isCacheableMethod && !hasCacheControl) {
 					// caches this page for up to a year
 					response.headers.append('CDN-Cache-Control', 'public, max-age=31536000, must-revalidate');
 				}


.changeset/dry-rules-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+only cache GET/HEAD requests with cacheOnDemandPages
```
alexanderniebuhr, line 5: Can we try to reword this changeset to follow our sematics, e.g. `Fixes ... to ..., Fixes ... which ...`

Some guidance questions, not mandatory to answer all, but helpful.
What does this change do?
How does this change do it?
Why is this change necessary?

packages/netlify/src/ssr-function.ts
```diff
@@ -36,14 +36,16 @@ export const createExports = (manifest: SSRManifest, _args: Args) => {
 			}
 
 			if (integrationConfig.cacheOnDemandPages) {
+				const isCacheableMethod = ['GET', 'HEAD'].includes(request.method);
```
alexanderniebuhr, line 39: Are `GET`, `HEAD` the only cacheable methods?

alexanderniebuhr (commented):
Thank you for this fix ❤️ 
I added some comments, otherwise I'm happy with the code.

.changeset/dry-rules-bathe.md
Skn0tt: Made an update in https://github.com/withastro/adapters/pull/127/commits/277fc09d372392109850c29e8c4691df93a1602c, how do you like it?

packages/netlify/src/ssr-function.ts
Skn0tt: They're not the only cacheable ones, but the ones that most caching proxies cache by default. Caching a POST request can be a footgun because they tend to have side effects, same goes for PUT or DELETE requests. GET and HEAD are the only methods where caching is mostly safe.

alexanderniebuhr (approved):

florian-lefebvre (approved):

MoustaphaDev (approved):

.changeset/dry-rules-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Fixes cacheOnDemandPages setting to only cache GET/HEAD requests by default
```
dreyfus92, line 5: Just a tiny suggestion, let me know what you guys think 😁
```suggestion
Enhanced caching behavior for the `cacheOnDemandPages` setting. Now, it explicitly caches only GET and HEAD requests by default.
```

alexanderniebuhr: Thanks for the suggestion, I think it is good to add some more details, like you did. However we as maintainers unofficially agreed to use the present tense, so I suggest:

Enhances caching behavior for the `cacheOnDemandPages` setting. Now, it explicitly caches only GET and HEAD requests by default.

sarah11918, line 5:
```suggestion
Updates the behavior of the `cacheOnDemandPages` setting to only cache GET/HEAD requests by default
```
Maybe a slight preference for "updates the behaviour" as presenting it more of a "thing you might want to check if you have this enabled because it changed" vs just saying "fix" (which more implies something that was broken before should now work, not "we changed how it worked").

Just a thought! Up to you! Either way, I would use code syntax for the property!

sarah11918 (commented):
Left a suggestion for your consideration!

alexanderniebuhr:
@Skn0tt Both suggestions for the changeset are fine for me too. Pick whichever you like more. But I do agree we should make the property (cacheOnDemandPages) using code syntax!

lilnasy (approved):
Looks great to me!

.changeset/dry-rules-bathe.md
VoxelMC, line 5: Since I'm here, I'll throw my hat into the ring! I agree with "updates", as nothing broke before!

```suggestion
Updates the default behavior of `cacheOnDemandPages` to cache only GET/HEAD requests 
```
Or you can place "default" back at the end. Up to you!

Skn0tt:
Both suggestions read fine to me, but I liked Sarah's one a tad better! Applied it :D

dreyfus92 (approved):
LGTM, 🚢 it.

PR #130, opened 2024-01-14
fix: updates to use the new app.render call signature

asdfjkalsdfla:
## Changes
- Updates the app.redner call to use the latest signature

## Testing
- runs the existing unit test

### DIFF
--- a/.changeset/curly-oranges-end.md
+++ b/.changeset/curly-oranges-end.md
@@ -1,6 +1,6 @@
 ---
-"@astrojs/cloudflare": minor
-"@astrojs/netlify": minor
+"@astrojs/cloudflare": major
+"@astrojs/netlify": major
 ---
 
 Updates the internals of the integration to support Astro 4.0. See this [upstream pull request](https://github.com/withastro/astro/pull/9199) for additional details. **Warning:** Make sure to upgrade your Astro version to `>4.0` as `3.0` is no longer supported.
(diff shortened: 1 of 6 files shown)


alexanderniebuhr:
@asdfjkalsdfla Thank you for the contribution ❤️ 
We'll wait for some other reviews, but it should be good to go.

alexanderniebuhr (approved):

.changeset/curly-oranges-end.md
```diff
@@ -0,0 +1,6 @@
+---
+"@astrojs/cloudflare": minor
+"@astrojs/netlify": minor
+---
+
+Updates internal use of `app.render()` to use new signature
```
sarah11918, line 6: If I'm not mistaken, this is the upstream PR with similar changes? https://github.com/withastro/astro/pull/9199/

I'd take a look at those changesets and see if any of those comments are relevant here. Notice how the changeset for the other adapters specifically mentions that you need to upgrade to Astro 4.0. Would that be the same here?

I might also include, since this major change is not happening in the same PR this time, linking back to that PR saying that it is upgrading for the Astro 4.0 changes made there.

sarah11918 (commented):
Thanks @asdfjkalsdfla for this contribution! @alexanderniebuhr asked me to review the changeset for our description of these changes, and so I've left a comment along with reference to what we did re: advice to users when the other adapters were updated for this internal change.

I hope this gives some helpful additional guidance here!

.changeset/curly-oranges-end.md
alexanderniebuhr: @sarah11918 Thanks for the comment, makes total sense to me. And you are right!

I wrote the changeset, so I'll update it!

PR #132, opened 2024-01-18
updates Astro core version

alexanderniebuhr:
## Changes

- bumps Astro versions
- needed for pending major release

### DIFF
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -45,14 +45,14 @@
     "tiny-glob": "^0.2.9"
   },
   "peerDependencies": {
-    "astro": "^4.0.0"
+    "astro": "^4.2.0"
   },
   "devDependencies": {
     "execa": "^8.0.1",
     "fast-glob": "^3.3.1",
     "@types/iarna__toml": "^2.0.2",
     "strip-ansi": "^7.1.0",
-    "astro": "^4.0.0",
+    "astro": "^4.2.0",
     "chai": "^4.3.10",
     "cheerio": "1.0.0-rc.12",
     "mocha": "^10.2.0",
--- a/.changeset/curly-oranges-end.md
+++ b/.changeset/curly-oranges-end.md
@@ -1,6 +1,5 @@
 ---
 "@astrojs/cloudflare": major
-"@astrojs/netlify": major
 ---
 
-Updates the internals of the integration to support Astro 4.0. See this [upstream pull request](https://github.com/withastro/astro/pull/9199) for additional details. **Warning:** Make sure to upgrade your Astro version to `>4.0` as `3.0` is no longer supported.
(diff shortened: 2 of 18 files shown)


alexanderniebuhr:
* `main`
  * **PR #132** <a href="https://app.graphite.dev/github/pr/withastro/adapters/132?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/withastro/adapters/132?utm_source=stack-comment).

ematipico (approved):

packages/cloudflare/package.json
```diff
@@ -45,14 +45,14 @@
     "tiny-glob": "^0.2.9"
   },
   "peerDependencies": {
-    "astro": "^4.0.0"
+    "astro": "^4.2.0"
```
Skn0tt, line 48: I'm curious, what change in the adapter makes it incompatible with `4.0`? I saw the change to `render`, but don't see how that requires an upgrade here. Or is this more of a nudge to keep users on current versions?

ematipico: Good point. This bump **isn't** needed if the adapters don't use the new APIs provided by `astro/app`

Skn0tt: You refer to https://github.com/withastro/astro/pull/9661, right? Is there anything else that's new in 4.2 that's useful to adapters?

Skn0tt: Those quality-of-life additions are definitely nice and useful, but I'm worried about raising the peer dependency so fast here. What about we wait for a couple weeks with using them, to give users time to upgrade?

ematipico: Yeah I don't mind waiting a bit :)

alexanderniebuhr: The point why I bumped the peer dependency to `4.2`, was that we would already be ready for the new features if we use them in a couple of weeks. Otherwise we would need to do another major release in two weeks or so. If that is fine for everyone, we can also take the approach of two major releases :)

Skn0tt: Oh no, also don't want us to do two major releases when it could be one. If we don't bump the peer dependency here, why do we need to cut a major release?

alexanderniebuhr: In theory it is a breaking change if we bump peer dependency versions, the user would need to do manual changes to keep the site compatible. We always handled it in the past like this, IIRC.

On the other hand, we only have to do another major release once we use the new features of 4.2, and we don't know yet if that would be weeks or months.

alexanderniebuhr: Yeah that make sense, and is fine for me. I'll set the peer dependencies for  the netlify adapter to `^4.0.0`, and once we start using the new features in the netlify adapter we can cut another major release.

_For Cloudflare I'll already use this release, because I'm planning to add the new features next week._

Issue #133, opened 2024-01-18
Netlify: Host fails to serve 404 page on failing routes when adapter is in use.


regisphilibert:
### Astro Info

```block
Astro                    v4.2.1
Node                     v21.1.0
System                   macOS (arm64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/netlify
Integrations             none
```

### Describe the Bug

The 404 logic of Netlify is broken on a deployed hybrid project. Netlify fails to serve the 404 page (Netlify's or project's own 404.html). See: https://imaginative-pastelito-41792a.netlify.app/notfound

If one tries to enforce the 404 logic by manually adding a redirect rule:

```
# _redirect
/* /404.html 404
```

Then SSR routes (not static) will all land on the provided 404 file.

See https://imaginative-pastelito-41792a.netlify.app/ deployed from minimal repo.

### What's the expected result?

When route is not found Netlify should serve its own 404 page or the project's own 404.html

### Link to Minimal Reproducible Example

https://github.com/regisphilibert/astro-netlify-404-repro

### Participation

- [ ] I am willing to submit a pull request for this issue.

Skn0tt:
Hey Regis! Thanks for providing the repro. For any page where there's no static asset generated at buildtime, the Netlify adapter is configured to fall back to SSR. SSR seems to render the blank 404 response then. I'd assume that it incorporates `pages/404.astro` when doing that. It doesn't seem like your repro contains a `404.astro` page, what happens if you add it?

alexanderniebuhr:
Not sure if that is relevant here, but for Cloudflare there is SPA mode, for which a pre-rendered 404 is needed. SSR 404 pages do not work for Cloudflare and will also return a blank page. Maybe this is related.

regisphilibert:
> For any page where there's no static asset generated at buildtime, the Netlify adapter is configured to fall back to SSR

I just realized some of our site would have the SSR functions invoked when someone visited a static route on that project. At first I couldn't understand what was going on since we only have one ssr route and it's only used by CMS editors.

But it turns out, we had a meta in the <head> pointing to local missing files. So the browser would request a 404 on every visit and an ssr function would be invoked. So as you said, whenever someone (or the browser/robots etc...) will hit a 404, the `ssr` function will be triggered. 

[...]

PR #143, opened 2024-01-29
fix(netlify): pre-rendered 404 pages aren't shown

Skn0tt:
## Changes

Adds tests around 404 pages. Fixes prerendered 404 pages by embedding them into the SSR Handler and returning it when no route matches. There's other solutions to this, as we're discussing in https://github.com/withastro/adapters/issues/133, but this is a quick & easy fix to get 404 pages functioning.

## Testing

added two unit tests.

## Docs

It's a fix, so no docs needed. Related to https://github.com/withastro/adapters/issues/133.

### DIFF
--- a/.changeset/serious-kids-suffer.md
+++ b/.changeset/serious-kids-suffer.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Fixes a bug in the Netlify Adapter where prerendered 404.astro pages weren't shown
--- a/.changeset/serious-kids-suffer.md
+++ b/.changeset/serious-kids-suffer.md
@@ -2,4 +2,4 @@
 '@astrojs/netlify': patch
 ---
 
-Fixes a bug in the Netlify Adapter where prerendered 404.astro pages weren't shown
+Fixes a bug in the Netlify Adapter where prerendered 404.astro pages weren't shown on hybrid/server deployments.
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -105,13 +105,14 @@ export default function netlifyIntegration(
 		}
 	}
 
-	async function writeSSRFunction() {
+	async function writeSSRFunction(notFoundContent?: string) {
 		await writeFile(
 			new URL('./ssr.mjs', ssrOutputDir()),
 			`
 				import createSSRHandler from './entry.mjs';
 				export default createSSRHandler(${JSON.stringify({
 					cacheOnDemandPages: Boolean(integrationConfig?.cacheOnDemandPages),
+					notFoundContent,
 				})});
 				export const config = { name: "Astro SSR", generator: "@astrojs/netlify@${packageVersion}", path: "/*", preferStatic: true };
 			`
@@ -300,7 +301,11 @@ export default function netlifyIntegration(
 				logger.info('Emitted _redirects');
 
 				if (_config.output !== 'static') {
-					await writeSSRFunction();
+					let notFoundContent = undefined;
+					try {
+						notFoundContent = await readFile(new URL('./404.html', dir), 'utf8');
+					} catch {}
+					await writeSSRFunction(notFoundContent);
 					logger.info('Generated SSR Function');
 				}
 
--- a/packages/netlify/src/ssr-function.ts
+++ b/packages/netlify/src/ssr-function.ts
@@ -13,11 +13,14 @@ const clientAddressSymbol = Symbol.for('astro.clientAddress');
 export const createExports = (manifest: SSRManifest, _args: Args) => {
 	const app = new App(manifest);
 
-	function createHandler(integrationConfig: { cacheOnDemandPages: boolean }) {
+	function createHandler(integrationConfig: { cacheOnDemandPages: boolean, notFoundContent?: string }) {
 		return async function handler(request: Request, context: Context) {
 			const routeData = app.match(request);
-			Reflect.set(request, clientAddressSymbol, context.ip);
+			if (!routeData && typeof integrationConfig.notFoundContent !== 'undefined') {
+				return new Response(integrationConfig.notFoundContent, { status: 404 });
+			}
 
+			Reflect.set(request, clientAddressSymbol, context.ip);
 			let locals: Record<string, unknown> = {};
 
 			const astroLocalsHeader = request.headers.get('x-astro-locals');
--- a/packages/netlify/test/functions/cookies.test.js
+++ b/packages/netlify/test/functions/cookies.test.js
@@ -23,4 +23,24 @@ describe('Cookies', () => {
 		expect(resp.headers.get('location')).to.equal('/');
 		expect(resp.headers.getSetCookie()).to.eql(['foo=foo; HttpOnly', 'bar=bar; HttpOnly']);
 	});
+
+	it("renders dynamic 404 page", async () => {
+		const entryURL = new URL(
+			'./fixtures/cookies/.netlify/functions-internal/ssr/ssr.mjs',
+			import.meta.url
+		);
+		const { default: handler } = await import(entryURL);
+		const resp = await handler(
+			new Request('http://example.com/nonexistant-page', {
+				headers: {
+					"x-test": "bar"
+				}
+			}),
+			{}
+		);
+		expect(resp.status).to.equal(404);
+		const text = await resp.text()
+		expect(text).to.contain("This is my custom 404 page");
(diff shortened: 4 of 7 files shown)


packages/netlify/src/index.ts
```diff
@@ -300,7 +301,11 @@ export default function netlifyIntegration(
 				logger.info('Emitted _redirects');
 
 				if (_config.output !== 'static') {
-					await writeSSRFunction();
+					let notFoundContent = undefined;
+					try {
+						notFoundContent = await readFile(new URL('./404.html', dir), 'utf8');
```
lilnasy, line 306: filesystem check would work reliably. That said, do you think we should prefer checking the routes object?

https://github.com/withastro/astro/pull/9648/files#diff-071ccfbdfd7b9bc803219179f88ae0bdb227a399bc4a481b7831c82b801d7a17R298

packages/netlify/src/ssr-function.ts
```diff
@@ -8,15 +8,15 @@ applyPolyfills();
 // biome-ignore lint/complexity/noBannedTypes: safe to use in this case
 export type Args = {};
 
-const clientAddressSymbol = Symbol.for('astro.clientAddress');
-
 export const createExports = (manifest: SSRManifest, _args: Args) => {
 	const app = new App(manifest);
 
-	function createHandler(integrationConfig: { cacheOnDemandPages: boolean }) {
+	function createHandler(integrationConfig: { cacheOnDemandPages: boolean, notFoundContent?: string }) {
```
lilnasy, line 14: So it seems like the approach is to make the serverless function serve the static file by inlining 404.html into the code.

How feasible would it be to make the serverless function act only on certain routes?

This is how it's done in the vercel adapter: https://github.com/withastro/astro/pull/9648/files#diff-071ccfbdfd7b9bc803219179f88ae0bdb227a399bc4a481b7831c82b801d7a17R290-R296

lilnasy: This would address the (admittedly small) billing concern.

Skn0tt: That would require us to generate a `_redirects` file, which opens up some complexity that i'd prefer not to have. The single benefit from that, as I understand it, would be that there's no function execution for 404 pages. I'm unsure how valuable that really is, and if we do that change i'd prefer to do it in a separate PR in the future.

packages/netlify/src/index.ts
lilnasy: I am happy as long as the users get an improvement, which this PR already provides. Just meant to point out that it is an option.

As for preferences, I am of the opinion it should be upto the more active contributor, which would be you.

alexanderniebuhr: I don't have any opinion here, fine with both approaches.

packages/netlify/src/ssr-function.ts
```diff
@@ -27,13 +27,7 @@ export const createExports = (manifest: SSRManifest, _args: Args) => {
...
 			locals.netlify = { context };
 
-			const response = await app.render(request, { routeData, locals });
-
-			if (app.setCookieHeaders) {
-				for (const setCookieHeader of app.setCookieHeaders(response)) {
-					response.headers.append('Set-Cookie', setCookieHeader);
-				}
-			}
+			const response = await app.render(request, { routeData, locals, addCookieHeader: true, clientAddress: context.ip });
```
alexanderniebuhr, line 30: nit: This change seems unrelated to this PR. I would personally prefer to separate it.

alexanderniebuhr (commented):
If I understand this correctly, only a `pre-rendered` 404 page will be embedded in the SSR function, correct?

If so, I don't get why we need this, my understanding from the discussion is that static assets are served before the SSR function, in that case a `pre-rendered` 404 page is a static asset (`html` file). Maybe someone could explain this again to me?

We would also like to get docs to look over the changeset, once we have the technical things aligned.

packages/netlify/src/ssr-function.ts
Skn0tt: oof, that commit was still in my working copy and I didn't notice it! will remove it.

Skn0tt:
> If so, I don't get why we need this, my understanding from the discussion is that static assets are served before the SSR function, in that case a pre-rendered 404 page is a static asset (html file). Maybe someone could explain this again to me?

Static assets are served before the SSR, that is true. That means if you go to `/404`, the 404 page will correctly be shown already without this PR. But if you go to any other non-existant page, like `/products/i-dont-exist`, there won't be a static asset on that route! So because the SSR function is configured as the fallback, it will be called, and now it needs to serve the 404 page. We could also serve a redirect to the 404 page, but that would be slightly different behaviour.

alexanderniebuhr:
@Skn0tt  Ohhh, so Netlify does not fallback to `/404` if a `404.html` static asset exists in the deployment.

That's a different behaviour from Cloudflare, but then this PR makes sense to me.

In conclusion I guess Cloudflare has the reverse issue _(which we handle by the `_routes.json`)_, if you don't have a `404.html` but also don't route requests to the SSR function, Cloudflare will return an empty response. Which can't happen with Netlify, because the SSR function is the fallback.

Skn0tt:
> Ohhh, so Netlify does not fallback to /404 if a 404.html static asset exists in the deployment. I would expect that..

Well, normally it does, but the Adapter explicitly sets the SSR function to serve all routes except for the ones with static assets:

https://github.com/withastro/adapters/blob/3854812c9bd3558857bb9084668925835dcf4ee0/packages/netlify/src/index.ts#L116

(see the `path: /*` and `preferStatic: true` options)

> Which can't happen with Netlify, because the SSR function is the fallback.

And that's exactly the reason why I prefer to do it the way around that I put here, even if it means that there's a function execution on every 404 :D

alexanderniebuhr:
Got it. Thanks!
I already asked docs to get some feedback for the changeset.

.changeset/serious-kids-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Fixes a bug in the Netlify Adapter where prerendered 404.astro pages weren't shown
```
sarah11918, line 5: Would love for this to have a *tiny* bit more context, if it's a simple one liner...

> Fixes a bug in the Netlify Adapter where prerendered `404.astro` pages weren't displayed when/if...

If you can't answer that sentence in a few words, then it's fine! This feels like a bug that if people had it, they knew about it.  :smile: 

(If this didn't *always* happen, but the description of when it did is complicated or "it depends", you could say "where sometimes prerendered..." but not blocking.)

sarah11918 (commented):
Docs review, as requested! 🫡

.changeset/serious-kids-suffer.md
alexanderniebuhr: @Skn0tt can you update the changeset?

Skn0tt: I do! I'm sorry, got lost in a bunch of other work.

Skn0tt:
```suggestion
Fixes a bug in the Netlify Adapter where prerendered 404.astro pages weren't shown on hybrid/server deployments.
```

packages/netlify/src/ssr-function.ts
alexanderniebuhr: @Skn0tt do you still want to remove the commit? We can also keep it in the PR if you want.

Skn0tt: removed it!

.changeset/serious-kids-suffer.md
Skn0tt, line 1: oh wow, this filename is quite dire 😨

alexanderniebuhr:
@Skn0tt let me know when you're ready. I would merge it later in the evening :)

Skn0tt:
Ready, feel free to merge!

alexanderniebuhr (approved):

jamiemill:
My custom 404 page is showing raw html if I invoke it by going to a missing path like `/foo`, and is served as text/plain.

If I visit `/404` directly it's rendered as HTML.

Could this be related to this change?

Skn0tt:
Uh-oh! Yes, that's related to this change. We're missing the content type here: https://github.com/withastro/adapters/blob/0a87e38eae63c043f053aea25a34bb81b6723895/packages/netlify/src/ssr-function.ts#L25

Skn0tt:
here's the fix: https://github.com/withastro/adapters/pull/162

PR #152, opened 2024-02-05
feat(netlify): verification for edge middleware

lilnasy:
## Changes
- Ensures that when `edgeMiddleware` is enabled, only the edge middleware can call the internal serverless function.

## Testing
Manual

## Docs
- https://github.com/withastro/docs/pull/6786

### DIFF
--- a/.changeset/odd-shoes-talk.md
+++ b/.changeset/odd-shoes-talk.md
@@ -1,0 +1,7 @@
+---
+'@astrojs/netlify': minor
+---
+
+Implements verification for edge middleware. This is a security measure to ensure that your serverless functions are only ever called by your edge middleware and not by a third party.
+
+When `edgeMiddleware` is enabled, the serverless function will now respond with `403 Forbidden` for requests that are not verified to have come from the generated edge middleware. No user action is necessary.
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -6,6 +6,7 @@ import type { AstroConfig, AstroIntegration, RouteData } from 'astro';
 import { AstroError } from 'astro/errors';
 import { build } from 'esbuild';
 import { appendFile, mkdir, readFile, rm, writeFile } from 'fs/promises';
+import type { Args } from "./ssr-function.js"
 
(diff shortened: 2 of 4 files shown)


lilnasy:
!preview edge-middleware-verification

Skn0tt (changes requested):
This is a risky change, and it's not obvious to me what attack vector this protects against. Could you outline that, either here or via email if you feel like it's too sensitive?

lilnasy:
This is motivated by the clerk vulnerability.

- [Investigating a major security vulnerability with Clerk's Next.js integration](https://pilcrow.vercel.app/blog/clerk-nextjs-vulnerability)
- [Clerk January 2024 Vulnerability Update](https://clerk.com/changelog/2024-02-02)

Skn0tt:
Got it, thanks! That looks like a spooky vuln definitely o.O

My understanding of the PR is that you want to ensure that the SSR function can only be reached via the Middleware, if there's a middleware configured. I don't think that attack vector exists currently, because the middleware is configured to run on all requests (except for the requests to Image CDN and static assets):

https://github.com/withastro/adapters/blob/7a5c3b06a7d62357dbeaa0fa060e35632aee61a4/packages/netlify/src/index.ts#L148

So there's no way of circumventing the Middleware function. We're already ensuring this works as expected inside of the CDN, I don't think we need another check in the adapter on top of that. Am I missing something here?

lilnasy:
Yeah, looks like Netlify gives us a guard rail, which is great. It would be even better to have a safety net on top of it.

Skn0tt:
Your call! If the behaviour is the same across all adapters, then it's fine I guess.

lilnasy:
@Skn0tt I have added you to a private thread in the #netlify channel of the astro discord. You might want to take that into account as well.

Skn0tt (approved):

alexanderniebuhr (commented):
Pretty sure the changeset is good, but let's have docs take a look ;)

sarah11918 (approved):
Changeset looks great!!  :muscle:

PR #160, opened 2024-02-08
fix(netlify): consider main fields when bundling

lilnasy:
## Changes
- Updates astro 4.2.0 to 4.3.5 to see if builds fail because of `cssesc`. (Edit: womp womp)
- Fixes one of the two errors brought up in https://github.com/withastro/astro/issues/9880

## Testing
Existing tests should pass.

## Docs
Does not affect usage

### DIFF
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -52,7 +52,6 @@
     "@types/iarna__toml": "^2.0.2",
     "strip-ansi": "^7.1.0",
     "astro": "^4.3.5",
-    "chai": "^4.3.10",
     "cheerio": "1.0.0-rc.12",
     "wrangler": "^3.15.0",
     "@astrojs/test-utils": "workspace:*",
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -44,7 +44,7 @@
     "tiny-glob": "^0.2.9"
   },
   "peerDependencies": {
-    "astro": "^4.3.5"
+    "astro": "^4.2.0"
   },
   "devDependencies": {
     "execa": "^8.0.1",
--- a/packages/cloudflare/package.json
+++ b/packages/cloudflare/package.json
@@ -44,14 +44,15 @@
     "tiny-glob": "^0.2.9"
   },
   "peerDependencies": {
-    "astro": "^4.2.0"
+    "astro": "^4.3.5"
   },
   "devDependencies": {
     "execa": "^8.0.1",
     "fast-glob": "^3.3.1",
     "@types/iarna__toml": "^2.0.2",
     "strip-ansi": "^7.1.0",
-    "astro": "^4.2.0",
+    "astro": "^4.3.5",
+    "chai": "^4.3.10",
     "cheerio": "1.0.0-rc.12",
     "wrangler": "^3.15.0",
     "@astrojs/test-utils": "workspace:*",
--- a/packages/netlify/package.json
+++ b/packages/netlify/package.json
@@ -41,7 +41,7 @@
     "esbuild": "^0.19.5"
   },
   "peerDependencies": {
-    "astro": "^4.3.5"
+    "astro": "^4.2.0"
   },
   "devDependencies": {
     "@netlify/edge-functions": "^2.0.0",
--- a/packages/netlify/package.json
+++ b/packages/netlify/package.json
@@ -41,13 +41,13 @@
     "esbuild": "^0.19.5"
   },
   "peerDependencies": {
-    "astro": "^4.2.0"
+    "astro": "^4.3.5"
   },
   "devDependencies": {
     "@netlify/edge-functions": "^2.0.0",
     "@netlify/edge-handler-types": "^0.34.1",
     "@types/node": "^18.17.8",
-    "astro": "^4.2.0",
+    "astro": "^4.3.5",
     "chai": "^4.3.10",
     "chai-jest-snapshot": "^2.0.0",
     "cheerio": "1.0.0-rc.12",
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -159,6 +159,7 @@ export default function netlifyIntegration(
 			entryPoints: [fileURLToPath(new URL('./entry.mjs', middlewareOutputDir()))],
 			target: 'es2022',
 			platform: 'neutral',
+			mainFields: [ 'main' ],
 			outfile: fileURLToPath(new URL('./middleware.mjs', middlewareOutputDir())),
 			allowOverwrite: true,
 			format: 'esm',
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -159,7 +159,7 @@ export default function netlifyIntegration(
 			entryPoints: [fileURLToPath(new URL('./entry.mjs', middlewareOutputDir()))],
 			target: 'es2022',
 			platform: 'neutral',
-			mainFields: [ 'main' ],
+			mainFields: [ 'main', 'module' ],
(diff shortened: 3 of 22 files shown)


lilnasy:
!preview cssesc-fix

packages/cloudflare/package.json
```diff
@@ -45,14 +45,14 @@
     "tiny-glob": "^0.2.9"
   },
   "peerDependencies": {
-    "astro": "^4.2.0"
+    "astro": "^4.3.5"
```
natemoo-re, line 48: I believe a `minor` bump to the `peerDependency` range should also correspond with a `minor` bump to the package, but I don't see `@astrojs/cloudflare` included in the changesets.

packages/netlify/test/functions/edge-middleware.test.js
```diff
@@ -39,7 +39,7 @@ describe('Middleware', () => {
 			expect(contents.includes('"Hello world"')).to.be.false;
 		});
 
-		it('does not apply middleware during prerendering', async () => {
+		it.skip('does not apply middleware during prerendering', async () => {
```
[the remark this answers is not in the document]
natemoo-re: Ah yeah, this behavior is maybe confusing, but the current intention is for middleware to run on prerendered _and_ dynamically rendered requests.

We probably need a better way to control this from `astro` itself rather than having it be an adapter decision.

packages/netlify/src/index.ts
```diff
@@ -155,6 +155,7 @@ export default function netlifyIntegration(
 			entryPoints: [fileURLToPath(new URL('./entry.mjs', middlewareOutputDir()))],
 			target: 'es2022',
 			platform: 'neutral',
+			mainFields: [ 'main' ],
```
natemoo-re, line 158: I believe this would ignore packages with a valid `module` field, instead pulling in the code from `main`? I think we want this instead:

```suggestion
			mainFields: [ 'module', 'main' ],
```

packages/netlify/package.json
```diff
@@ -41,13 +41,13 @@
     "esbuild": "^0.19.5"
   },
   "peerDependencies": {
-    "astro": "^4.2.0"
+    "astro": "^4.3.5"
```
lilnasy, line 44:
```suggestion
    "astro": "^4.2.0"
```

packages/netlify/src/index.ts
lilnasy: I found out about this config when bluwy mentioned it. From what I can tell from esbuild docs ([Main Fields](https://esbuild.github.io/api/#main-fields)) , `"module"` is intended for code targeting browsers. Would we want esbuild to consider it for middleware?

bluwy: I think we should add the `module` field here. The esbuild docs mention that people tend to use `module` for browser-targetting envs, but it's not what `module` field was meant to be. They should use the `browser` field instead.

Skn0tt: There's more docs on this in the [Platform section](https://esbuild.github.io/api/#platform) of ESBuild docs:

> When the platform is set to neutral:
> ....
> The [main fields](https://esbuild.github.io/api/#main-fields) setting is empty by default. If you want to use npm-style packages, you will likely have to configure this to be something else such as `main` for the standard `main` field used by node.

That makes me think we should use `["main", "module"]`, in that order, because ESBuild will use the array ordering as the precedence order. We're bundling Edge Functions here, which run in an environment that's more similar to Node.js than it is to Browsers. So if `module` sometimes is used for browser builds, we should prefer the Node.js `main` build over that.

Having said that, I don't have too strong of an opinion on the ordering here. But I do agree that we should include both, yes!

packages/netlify/test/functions/edge-middleware.test.js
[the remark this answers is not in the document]
Skn0tt: Totally agree that this should be controlled from Astro itself - it's an essential piece of how rendering and middleware interact, and that shouldn't live in an adapter that's different for every provider.

When we did the rewrite of this plugin, the Astro docs weren't yet clear about the intended behaviour on this. @lilnasy you mention there was a bug fix, could you link me to more info on this?

My 2cts is that i'd like to remove the edgeMiddleware field from the Adapter options, and have Astro dictate the behaviour here. @natemoo-re you're referring to "the current intention" - is the behaviour around this in-flux? If yes, where can I read up on that so we make sure that the adapter behaves correctly?

The docs on this say "Middleware allows you to intercept requests and responses and inject behaviors dynamically every time a page or endpoint is about to be rendered.", which I find to be a bit fuzzy. Personally, I'd prefer the behaviour to be defined a little more explicit here. Is it "dynamically", or is it "on render"? "dynamically" would mean "on every request" to me, "on render" would mean "depends on your prerender option".

packages/netlify/src/index.ts
lilnasy: main then module sounds good to me, updating the suggestion to add it.

bluwy: Usually when both main and module exist, module should take precedence first, so I'm not really sure flipping that is a good idea.

packages/netlify/test/functions/edge-middleware.test.js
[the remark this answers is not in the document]
lilnasy: > My 2cts is that i'd like to remove the edgeMiddleware field from the Adapter options, and have Astro dictate the behaviour here.

That was the original implementation. We switched letting the adapter decide it in Astro 3.

I don't think we ever sat down to discuss pre-rendered pages with edgeMiddleware enabled, but even from the test case, it comes across as a bug that pages wouldn't have access to locals.

Also I should clarify, when I said this behavior isn't the adapter's to control, I meant that is currently the case - the netlify adapter did nothing to hide locals before, and nothing to add them since Astro 4.2.

packages/netlify/src/index.ts
lilnasy: Is it a common config?

bluwy: Yeah, it's common for module and main to be set together. And usually module comes first, then main. `main` is like the "universal fallback" that most toolings know, so having module take precedence and be the more specific alternative makes more sense.

Thanks for making the change!

lilnasy: Thanks. If it's common, I don't think we don't need reconsider it.

bluwy (approved):
LGTM after resolving the conflicts!

packages/cloudflare/package.json
```diff
@@ -51,7 +51,8 @@
     "fast-glob": "^3.3.1",
     "@types/iarna__toml": "^2.0.2",
     "strip-ansi": "^7.1.0",
-    "astro": "^4.2.0",
+    "astro": "^4.3.5",
+    "chai": "^4.3.10",
```
ematipico, line 55: I think this is a mistake

lilnasy: This came up in the conflict. I thought it was recently added to `main`. Was it removed?

alexanderniebuhr (changes requested):
I guess what @natemoo-re said about version bumps and missing changesets is still true. What is the planned action, revert those version bumps or make a release for both packages?

lilnasy:
@alexanderniebuhr they were reverted

alexanderniebuhr:
@lilnasy Sure. I still see Astro being bumped from `^4.2.0` to `^4.3.5` in both Adapters, but only a release for Netlify. While I guess it should be a working semver range, I guess we still should also cut a release for the Cloudflare adapter so that there are no issues, shouldn't we?

lilnasy:
Only devDependencies changed, which would be ignored for users.

alexanderniebuhr (approved):

alexanderniebuhr:
You're right, was looking at the test fixtures, but that doesn't bother users either...

PR #163, opened 2024-02-09
fix(netlify): replace `global.crypto` with `crypto`

OiYouYeahYou:
## Changes

This is a drive by change because `randomUUID` is not on the global version of `crypto` on my node version (`v18.18.2`). I'm uncertain if this breaks any compatibility for others and do not have pnpm to run and commands.

## Testing

Not testing done because this is a trivial change and don't have pnpm

## Docs

Trivial change, probably should be noted in changelog

### DIFF
--- a/.changeset/young-ravens-hammer.md
+++ b/.changeset/young-ravens-hammer.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/netlify": patch
+---
+
+Fixes an issue where using `global.crypto` failed for older Astro versions
--- a/.changeset/young-ravens-hammer.md
+++ b/.changeset/young-ravens-hammer.md
@@ -2,4 +2,4 @@
 "@astrojs/netlify": patch
 ---
 
-Fixes an issue where using `global.crypto` failed for older Astro versions
+Fixes an issue where some astro CLI commands failed with `crypto is not defined` on Astro 4.4.0 and earlier.
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -6,6 +6,7 @@ import type { AstroConfig, AstroIntegration, RouteData } from 'astro';
 import { AstroError } from 'astro/errors';
 import { build } from 'esbuild';
 import { appendFile, mkdir, readFile, rm, writeFile } from 'fs/promises';
+import { randomUUID } from 'crypto'
 import type { Args } from './ssr-function.js';
 
 const { version: packageVersion } = JSON.parse(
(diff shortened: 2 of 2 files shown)


alexanderniebuhr:
cc @Skn0tt, wdyt about this change?

Skn0tt:
Looks fine to me!

I don't think the global crypto polyfill will work, because this is about usage within the build, not within the runtime. And we only put that polyfill in place for the runtime.

bluwy:
The code I linked is used to apply the polyfill in build time.

Skn0tt:
Interesting, didn't know that! So why is `globalThis.crypto` undefined for @OiYouYeahYou?

lilnasy:
@OiYouYeahYou does it only happen with the `astro add` command?

https://github.com/withastro/astro/pull/10105

lilnasy:
I can only reproduce it with `astro add` and a few other commands (fixed by https://github.com/withastro/astro/pull/10105), but the dev command seems to be correctly polyfilled. Do you think something else about your project could be involved?

image

lilnasy (approved):
Looks great to me!

The PR to astro will enable the use of `crypto` and `File` in the config file for new versions. This PR can help with compatibility with older versions.

Could you add a patch changeset with `pnpm -w exec changeset`? It will show up in [CHANGELOG.md](https://github.com/withastro/adapters/blob/main/packages/netlify/CHANGELOG.md).

ematipico (approved):

alexanderniebuhr (approved):

.changeset/young-ravens-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@astrojs/netlify": patch
+---
+
+Fixes an issue where using `global.crypto` failed for older Astro versions
```
sarah11918, line 5: Can we give a version number here to be more specific? Something like:

> failed for Astro vx.x and earlier

(Assuming this fix works for *all* earlier Astro versions. Otherwise, specifying the range would be a nice addition.)

sarah11918 (approved):
Looks good to docs, but as mentioned in the comment, would love to be specific about version numbers if it's easy, but not a deal breaker!

(Especially if by chance it now works for *some* older versions, but not *all* of them, would be nice for people to know whether or not it should work for their version after this fix.)

.changeset/young-ravens-hammer.md
alexanderniebuhr: @lilnasy do you know which versions this applies to? I guess: `<v4.4.1`?

lilnasy: Yes, it was fixed in 4.4.1.

Issue #149, opened 2024-02-03
deduplication does not work


alexanderniebuhr:
### Describe the Bug

The deduplicatePatterns of `_routes.json` still doesn't work

### What's the expected result?

Correct deduplication of:

```
{
  "version": 1,
  "include": [
    "/",
    "/u/*",
    "/login",
    "/_image",
    "/sign-up",
    "/discover/*",
    "/freelancer",
    "/login/utils",
    "/api/auth/login",
    "/login/services",
    "/api/auth/logout",
    "/discover/*/utils"
  ],
  "exclude": []
}
```

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #137, opened 2024-01-19
fix(cloudflare): duplicate patterns in `_routes.json`

alexanderniebuhr:
## Changes

- closes #135
- closes #136

PR #171, opened 2024-02-21
fix(cloudflare): _routes.json deduplication

alexanderniebuhr:
## Changes

- this should fix the deduplication error we had for more complex route definitions
- this is a fix for v9 of the adapter, in v10 we will refactor the complete route generation logic _(so this has a limited time to live)_
- also reverting change in test file https://github.com/withastro/adapters/pull/137/files#diff-d3130bfe10263d21a437f7e826fc4594ffc77e5cc37be1b09b10f5553b08c3ed
- closes https://github.com/withastro/adapters/issues/149

## Testing

- I personally don't think it is worth to add a complete fixture for such edge-case, however we can if anyone has strong opinions
- Let's investigate if we can add unit tests instead which test only util functions

### DIFF
--- a/.changeset/sharp-cameras-rescue.md
+++ b/.changeset/sharp-cameras-rescue.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fixes automatic deduplication of `_routes.json`, which had more complex project structures
+Fixes an error with automatic deduplication of `_routes.json` for more complex project structures
--- a/.changeset/sharp-cameras-rescue.md
+++ b/.changeset/sharp-cameras-rescue.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes automatic deduplication of `_routes.json`, which had more complex project structures
--- a/packages/cloudflare/src/utils/deduplicatePatterns.ts
+++ b/packages/cloudflare/src/utils/deduplicatePatterns.ts
@@ -13,7 +13,7 @@ export function deduplicatePatterns(patterns: string[]) {
 	const uniquePatterns = [...new Set(patterns)];
 	for (const pattern of uniquePatterns) {
 		if (pattern.endsWith('*')) {
-			openPatterns.push(new RegExp(`^${pattern.replace(/(\*\/)*\*$/g, '[^*\n]*$')}`));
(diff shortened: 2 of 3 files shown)


alexanderniebuhr:
* **#171** <a href="https://app.graphite.dev/github/pr/withastro/adapters/171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/withastro/adapters/171?utm_source=stack-comment).

packages/cloudflare/src/utils/deduplicatePatterns.ts
```diff
@@ -13,7 +13,7 @@ export function deduplicatePatterns(patterns: string[]) {
 	const uniquePatterns = [...new Set(patterns)];
 	for (const pattern of uniquePatterns) {
 		if (pattern.endsWith('*')) {
-			openPatterns.push(new RegExp(`^${pattern.replace(/(\*\/)*\*$/g, '[^*\n]*$')}`));
+			openPatterns.push(new RegExp(`^${pattern.replace(/(\*\/)*\*$/g, '(?=.{2,}).+[^*\n]$')}`));
```
ematipico, line 16: What's the actual fix, though? The PR doesn't explain it

.changeset/sharp-cameras-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes automatic deduplication of `_routes.json`, which had more complex project structures
```
sarah11918, line 5:
```suggestion
Fixes an error with automatic deduplication of `_routes.json` for more complex project structures
```

sarah11918 (commented):
Maybe something like this, if it's accurate?

ematipico (approved):

MehrdadKhnzd:
Hi! It still doesn't work, I still get the error I used to get:
`Error: Failed to publish your Function. Got error: Error 8000057: Overlapping rules in _routes.json are not allowed.`

MehrdadKhnzd:
@alexanderniebuhr The problem is that I couldn't. I tried to reproduce it two times, but I wasn't successful when I partially recreated some of the conflicting structures of my project. I even thought it might be due to some other added plugins, but that wasn't the case. I might give it another try with a complete replication of my file structure to see how it goes, but it will be really time-consuming. Do you know when v10 is going to be ready?

PR #197, opened 2024-03-21
removes esbuild step

alexanderniebuhr:
## Changes

- removes the additional esbuild build step we had
- this allows to use the same build output from Astro's vite `astro build` for Cloudflare, than any other adapter
- **only difference** that we set `vite.ssr.noExternal = true;`, because Cloudflare doesn't have the concept of a filessystem, therefore won't be able to have a `node_modules` folder
- this will lead to a `./dist/_worker.js/` output **directory**, which is recommended by Cloudflare for framework authors and allows use to use lazy loading of pages in Cloudflare, if our output is build that way

## Testing

- fixed some tests
- ignored failing tests :re `_routes.json` generation, since the generation logic is overhauled in a following PR

## Docs

⚠️ needs v10 major doc update PR

PR #194, opened 2024-03-21
inits major upgrade

alexanderniebuhr:
## Changes

- updates dependencies
- adds `.nvmrc` for easier node version management
- fix `_test-utils.js` due to breaking change in dependency update
- increase testing timeout

## Testing

- nothing added

## Docs

- ⚠️ needs v10 major doc update PR

Issue #193, opened 2024-03-21
When `build.format: "file"`, `_routes.json` is not configured correctly and SSR is preferred over SSG.


totto2727:
### Astro Info

```block
Astro                    v4.4.15
Node                     v20.11.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/svelte
```

### Describe the Bug

The output of Cloudflare Adaptor,` dist/_routes.json`, has a different configuration depending on the value of build.format in `astro.config.mjs`.

In particular, the `exclude` in `build.format: "file"` does not seem to be set correctly.

- `/pages/ssg/index.astro`: prerender = true
- `/pages/ssg/[ssr]/index.astro`: prerender = false

In `build.format: "directory"`, `/pages/ssg/index.astro` takes precedence over `/pages/ssg/[ssr]/index.astro`.

However, with `build.format: "file"`, `/pages/ssg/[ssr]/index.astro` takes precedence and a 404 is returned when trying to access `/pages/ssg/index.astro`.

This is because `exclude` does not include "/ssg/" and SSR with "/ssg/*" in `include` are preferred.

This is not surprising, since the SSR script does not have a corresponding process for `/pages/ssg/index.astro`, but the difference in behavior is annoying.

### What's the expected result?

This is also true for `mode: "hybrid"`.

# `dist/_route.json` outputs

### expected

```json
{
  "version": 1,
  "include": [
    "/ssg/*",
    "/_image"
  ],
  "exclude": [
    "/ssg/",
    "/ssg.html"
  ]
}
```

### actual

```json
{
  "version": 1,
  "include": [
    "/ssg/*",
    "/_image"
  ],
  "exclude": []
}
```

## `mode: "server", build.format: "file"`

No problem.

### 

```json
{
  "version": 1,
  "include": [
    "/ssg/*",
    "/_image"
  ],
  "exclude": [
    "/ssg/",
    "/ssg/index.html"
  ]
}
```

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-3jxjgu?file=dist%2F_routes.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

PR #159, opened 2024-02-08
feat(cloudflare): major

alexanderniebuhr:
## Changes

- announced PR for Cloudflare Adapter Major v10
- closes #182 
- closes #180
- closes #78
- closes #153
- closes #134
- closes #193
- closes #202 

## Docs

- ⚠️ needs v10 major doc update PR
- https://github.com/withastro/docs/pull/7592

PR #199, opened 2024-03-21
refactors routes generation

alexanderniebuhr:
## Changes

- refactors the route generation logic
- uses a Trie like structure
- align with Cloudflare routing rules for wildcards `/x/*` -> a wildcard matches all, even more specific routes like `/x/y/*`, are not allowed due to duplicated rules
- `_routes.json` generation should be a best approach, meaning when reaching more than 100 entries (Cloudflare's) limit, requests should be execute and invoke SSR and fallback to a ASSET fetch behavior
- The new logic should be much more robust to exclude complete folders rather all files

## Testing

- need to add adapted route json test, but I'll do that later, before merging

## Docs

⚠️ needs v10 major doc update PR


alexanderniebuhr:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/withastro/adapters/199?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#199** <a href="https://app.graphite.dev/github/pr/withastro/adapters/199?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#198** <a href="https://app.graphite.dev/github/pr/withastro/adapters/198?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#197** <a href="https://app.graphite.dev/github/pr/withastro/adapters/197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196** <a href="https://app.graphite.dev/github/pr/withastro/adapters/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#195** <a href="https://app.graphite.dev/github/pr/withastro/adapters/195?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#194** <a href="https://app.graphite.dev/github/pr/withastro/adapters/194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#159** <a href="https://app.graphite.dev/github/pr/withastro/adapters/159?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @alexanderniebuhr and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/cloudflare/src/index.ts
```diff
@@ -1,34 +1,44 @@
...
-		/** Additional `include` patterns */
-		include?: string[];
-		/** Additional `exclude` patterns */
-		exclude?: string[];
+		/** Extend `_routes.json` */
+		extend: {
+			/** Paths which should be routed to the SSR function */
+			include?: {
+				/** Generally this is in pathname format, but does support wildcards, e.g. `/users`, `/products/*` */
+				pattern: string;
```
ematipico, line 34: What's the motivation for making this an array of objects instead of an array of strings (like it was before)? It isn't explained in the PR

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+import { existsSync } from 'node:fs';
+import { writeFile } from 'node:fs/promises';
+import { posix } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import glob from 'tiny-glob';
+import { removeLeadingForwardSlash, removeTrailingForwardSlash } from './assets.js';
+
+function prependForwardSlash(path: string) {
+	return path[0] === '/' ? path : `/${path}`;
+}
```
ematipico, line 12: You should use `@astrojs/internal-helpers`, so you use the same code that `astro` uses to handle paths

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+		const [, content] = dynamic ? /([^(]+)$/.exec(str) || [null, null] : [null, str];
+
+		if (!content || (dynamic && !/^(?:\.\.\.)?[\w$]+$/.test(content))) {
+			throw new Error('Parameter name must match /^[a-zA-Z0-9_$]+$/');
+		}
+
+		result.push({
+			content,
+			dynamic,
+			spread: dynamic && /^\.{3}.+$/.test(content),
```
ematipico, line 29: Don't put regex inside loops; it's really bad for performance. At least define them inside the function before the loop.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+		else pathSegments.push(segment.content);
+	}
+	return pathSegments;
+};
+
+class TrieNode {
+	children: Map<string, TrieNode> = new Map();
+	isEndOfPath = false;
+	hasWildcardChild = false;
+}
```
ematipico, line 52: It would be nice to have internal documentation explaining what this class does. You mentioned "Trie" in the PR description, but you assume that everybody knows that.

Having some reference somewhere is useful so another developer can also pick the knowledge base.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+			if (segment === '*') {
+				node.hasWildcardChild = true;
+				break;
+			}
+			if (!node.children.has(segment)) {
+				node.children.set(segment, new TrieNode());
+			}
+
+			// biome-ignore lint/style/noNonNullAssertion: <explanation>
+			node = node.children.get(segment)!;
```
ematipico, line 73: You turned off a rule for something that is unsafe...  without an explanation. Looking at the code, without some context, this is unsafe, because `children` could be empty indeed. 

You should provide more explanation or refactor the code to make it more robust.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		node.isEndOfPath = true;
+	}
+
+	private dfs(node: TrieNode, path: string[], allPaths: string[][]): void {
```
ematipico, line 79: nit: What does `dfs` mean? Probably something related to "Trie", but still, the name doesn't help.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+			allPaths.push([...path, '*']);
+			return;
+		}
+
+		if (node.isEndOfPath) {
+			allPaths.push([...path]);
+		}
+
+		for (const [segment, childNode] of node.children) {
+			this.dfs(childNode, [...path, segment], allPaths);
```
ematipico, line 90: nit: I am a bit skeptical about a recursion with inline mutations. But you also create a data structure that involves nesting... I am not sure it's the best approach unless there's something I am missing.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	getAllPaths(): string[][] {
+		const allPaths: string[][] = [];
+		this.dfs(this.root, [], allPaths);
+		return allPaths;
+	}
+}
+
+export default async function (
```
ematipico, line 101: I would export a named function instead of mixing default exports and named exports. This is usually bad for three-shaking and other stuff.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+	_config: AstroConfig,
+	routes: RouteData[],
+	pages: {
+		pathname: string;
+	}[],
+	redirects: RoutePart[][][],
```
ematipico, line 107: Can we type this differently? We loop over `redirects`, and then we pass it to the function `segmentsToCfSyntax`. This function doesn't accept `RoutePart[][]`, but it accepts another type: `RouteData['segments']`.

Maybe `RouteData['segments'][]` is more **in-line** with the whole refactor.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+) {
+	const includePaths: string[][] = [];
+	const excludePaths: string[][] = [];
+
+	let hasPrerendered404 = false;
+	for (const route of routes) {
+		const convertedPath = segmentsToCfSyntax(route.segments, _config);
+		if (route.pathname === '/404' && route.prerender === true) hasPrerendered404 = true;
+
+		if (route.type === 'page') if (route.prerender === false) includePaths.push(convertedPath);
```
ematipico, line 127: Use `switch (route.type)` and handle all the variants. This is a better strategy, because if and when astro will add a new variant, your code will fail and you will be forced to handle it.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	const assetsPath = segmentsToCfSyntax(
+		[
+			[{ content: _config.build.assets, dynamic: false, spread: false }],
+			[{ content: '', dynamic: true, spread: false }],
+		],
+		_config
+	);
+	excludePaths.push(assetsPath);
```
ematipico, line 173: Let's add a comment that explains what you're doing

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+
+	const pagefindPath = segmentsToCfSyntax(
+		[
+			[{ content: 'pagefind', dynamic: false, spread: false }],
+			[{ content: '', dynamic: true, spread: false }],
+		],
+		_config
+	);
```
ematipico, line 181: Let's add a comment that explains what you're doing

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+
+	const excludeTrie = new PathTrie();
+	for (const excludePath of excludePaths) {
+		excludeTrie.insert(excludePath);
+	}
+	const deduplicatedExcludePaths = excludeTrie.getAllPaths();
+
+	if (
+		!hasPrerendered404 ||
+		deduplicatedIncludePaths.length > 100 ||
```
ematipico, line 202: Two things:
- lone numbers aren't self-explanatory, it's usually better to assign them to constants so the name helps to give them some context
- comment with a link to the reference explaining the limit. That's a very important piece of information.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -0,0 +1,267 @@
...
+							.slice(0, 99)
+							.concat(excludeExtends?.map((entry) => entry.pattern) ?? []),
+					},
+					null,
+					2
+				),
+				'utf-8'
+			);
+		} catch (error) {
+			// TODO
```
ematipico, line 223: You should address this `TODO`

packages/cloudflare/src/utils/generate-routes-json.ts
alexanderniebuhr: I copied this from https://github.com/withastro/astro/blob/3776ecf0aa9e08a992d3ae76e90682fd04093721/packages/astro/src/core/routing/manifest/create.ts#L45-L70, because I wanted to reuse it, but it isn't exported from Astro. 

I would prefer to have it exported from Astro. I could also import it from `node_modules` manually, but that is also not great.

Anyways I did copied it before it was changed in Astro, so my version is already outdated, so for now I'll copy the new version.

packages/cloudflare/src/utils/generate-routes-json.ts
alexanderniebuhr: I think the `if` check in [L68-L70](https://github.com/withastro/adapters/pull/199/files#diff-5e7c378df233d3953841865f54333b6df6c68ee3861ddf1e7bf083c83e23739bR68-R70) should make sure that the map has the `segment`.

```
if (!node.children.has(segment)) {
	node.children.set(segment, new TrieNode());
}
```

packages/cloudflare/src/utils/generate-routes-json.ts
alexanderniebuhr: So we want to get all unique paths, but if there is a path with a wildcard `(*)`, we don't want to return any path that is already matched by this wildcard path.
The previous version was based on RegExp which did not catch all cases and it was rather hard to maintain. See all the issues for it below. I also added a input / output example below.

Two other approaches would be either a Trie-like structure or a map-based approach. The map-based approach is flat while the Trie-like structure is nested, as you already said. In the Trie-like structure each node in the graph represents a path segment.

**Map-based approach advantages:**
- Maps are built-in structures, making the implementation easier.
- Maps will have a smaller memory footprint, as it doesn't need to allocate memory for each path segment.
- Maps implementation can be more flexible for adding additional logic and rules, because the underlaying structure might not need a refactor.

**Map-based approach disadvantages:**
- Maps can have more error-prone wildcard handling which captures all cases, like precedence and overlaps between wildcard paths.
- Maps do have generally time complexity of O(1), however checking for wildcards will need additional logic which adds more time complexity _(nested for loops)_

**Trie-like structure advantages:**
- Tries naturally handle wildcards which can be represented as a node that captures all possible continuations of a path from that point.
- Tries allow for fast lookup. Time complexity is generally O(n)
- Tries excel for paths with prefix matching. They can easily find all paths that share a common prefix and therefore only return the necessary onces.

**Trie-like structure disadvantages:**
- Tries can consume more memory. Each node represents a path segment, so paths with many segments increase the memory footprint.
- Tries with logic needs custom implementation and can be more complex than built-in data structures.

I overall think a Trie-like structure gives more flexibility and the benefits outweigh the potential disadvantages, memory footprint shouldn't be a big factor, since this runs during `astro build` and not in runtime.

Issues with the RegExp approach:
- #180
- #78
- #193
- #202

The following is an example of input and expected output:

```
// Input

[
  [ '', '_image' ],
  [ '', '404' ],
  [ '', 'products', '*' ],
  [ '', 'products', '*' ],
  [ '', 'products', 'endpoint' ],
  [ '', 'products', '*', 'overview' ],
  [ '', 'users', '*', 'profile' ],
  [ '', 'users', 'settings', '*' ],
  [ '', 'users', '*', 'settings', '*' ]
]

// Output

[
  [ '', '_image' ],
  [ '', '404' ],
  [ '', 'products', '*' ],
  [ '', 'users', '*' ],
]
```

packages/cloudflare/src/index.ts
alexanderniebuhr: @ematipico or do you think it would be better to either have an array of strings or rename the `pattern` property to `path`?

packages/cloudflare/src/utils/generate-routes-json.ts
ematipico: awesome, then we should mention it in `<explanation>` :)

packages/cloudflare/src/utils/generate-routes-json.ts
ematipico: > I overall think a Trie-like structure gives more flexibility and the benefits outweigh the potential disadvantages, memory footprint shouldn't be a big factor, since this runs during astro build and not in runtime.

That's the part that concerns me. Build times **matter**. We're having reports where build times are very long. 

Anyway, I wouldn't expect this to be a problem in the long run; it's not like we're handling 500k paths at once. If that happens in the future, we might want to consider a different strategy.

If you already know that some users need to handle 500k+ paths at once, then I suggest a different strategy. You can still use a Trie-like strategy, but you could evaluate a flattening approach, where you first flatten the tree and then loop over them.

packages/cloudflare/src/index.ts
ematipico: Nope, the explanation totally makes sense! 

> or rename the pattern property to path?

If you expect to receive things like `/path/*`, then `pattern` or `glob` are better names. Maybe `pattern` fits better because I don't think you'll expect to receive actual globs e.g. `path/{foo, bar}`

packages/cloudflare/src/utils/generate-routes-json.ts
alexanderniebuhr: I agree with you. I would like to start with this implementation and reconsider it, once we receive user complaints.

Issue #209, opened 2024-03-29
Cloudflare v10.0.0 Breaks Solid.js Build


MehrdadKhnzd:
### Astro Info

```block
Astro                    v4.5.12
Node                     v20.11.1
System                   Linux (x64)
Package Manager          bun
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             unocss
                         @astrojs/solid-js
                         @swup/astro
                         @vite-pwa/astro-integration
```

### Describe the Bug

I wasn't using v9 because of the route generation bug. Today I finally migrated to v10, but it broke something else.
Whenever I build the project, the server is built successfully, but when it wants to build the client (I'm using Solid.js), it throws a strange error:
```
Error when using sourcemap for reporting an error: Can't resolve original location of error.
[vite] ✓ 148 modules transformed.
[ERROR] [vite] x Build failed in ...
"template" is not exported by "node_modules/solid-js/web/dist/server.js", imported by ...
file: ...
1: import { template as _$template } from "solid-js/web";
            ^
...
```
As you see, this error is from the compiled Solid.js code, not my source code. I tried changing the order of integrations and removing other integrations to see if it solves the issue, but it didn't. The only thing that solved the issue (on a surface level) was to remove `solid()` integration itself from the Astro config. Surprisingly, the build was successful and it didn't throw any errors, but obviously, none of the pages work. When I roll Cloudflare adapter back to 8.1.0, it works and builds the project normally.

### What's the expected result?

To properly build the project without any errors.

### Link to Minimal Reproducible Example

Unfortunately, similar to the routes problem I had with v9, I couldn't reproduce it on a minimal setup given all the configurations I added.

### Participation

- [ ] I am willing to submit a pull request for this issue.

alexanderniebuhr:
Thanks for reporting this! 
I would love to investigate this and debug this for you! But I can't reproduce it. We need to figure out why this happens! I really want you to have a good experience on v10.

Can you share the project where it occurs? Also the error states some
template import from solid, are you doing something like that with solid.

alexanderniebuhr:
One quick idea, could it be that your solid packages are not updated to the latest version?

AkashRajpurohit:
@alexanderniebuhr facing the same issue, Here are the solidjs related packages

- `"@astrojs/solid-js": "^4.0.1"`
- `"solid-js": "^1.8.16",`

Both seems to be on the latest version, however upgrading cloudflare adapter to v10 breaks the build for me as well.

[...]

PR #212, opened 2024-03-29
fixes solid

alexanderniebuhr:
## Changes

- closes https://github.com/withastro/adapters/issues/209

## Testing

- manual tested + existing solid test

## Docs

- bug fix, no docs needed

### DIFF
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -188,6 +188,10 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						...vite.define,
 					};
 				}
+				// we thought that vite config inside `if (target === 'server')` would not apply for client
+				// but it seems like the same `vite` reference is used for both
+				// so we need to reset the previous conflicting setting
+				// in the future we should look into a more robust solution
 				if (target === 'client') {
 					vite.resolve ||= {};
 					vite.resolve.conditions ||= [];
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -152,14 +152,6 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							find: 'react-dom/server',
 							replacement: 'react-dom/server.browser',
 						},
-						{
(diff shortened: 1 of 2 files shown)


alexanderniebuhr:
!preview cf-10-solid

packages/cloudflare/src/index.ts
```diff
@@ -191,6 +186,13 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						...vite.define,
 					};
 				}
+				if (target === 'client') {
+					vite.resolve ||= {};
+					vite.resolve.conditions ||= [];
+					vite.resolve.conditions = vite.resolve.conditions.filter(
+						(c) => c !== 'workerd' && c !== 'worker'
+					);
+				}
```
ematipico, line 195: It's always good to leave comments that explain these... workaround?

ematipico (approved):

packages/cloudflare/src/index.ts
```diff
@@ -170,6 +162,9 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						}
 					}
 
+					vite.resolve.conditions ||= [];
+					vite.resolve.conditions.push('workerd', 'worker');
```
alexanderniebuhr, line 166:
```suggestion
					// We need those conditions, previous these conditions where applied at the esbuild step which we removed
					// https://github.com/withastro/astro/pull/7092
					vite.resolve.conditions.push('workerd', 'worker');
```

packages/cloudflare/src/index.ts
```diff
@@ -191,6 +188,13 @@ export default function createIntegration(args?: Options): AstroIntegration {
 						...vite.define,
 					};
 				}
+				if (target === 'client') {
```
alexanderniebuhr, line 191:
```suggestion
				// we thought that vite config inside `if (target === 'server')` would not apply for client
				// but it seems like the same `vite` reference is used for both
				// so we need to reset the previous conflicting setting
				// in the future we should look into a more robust solution
				if (target === 'client') {
```

PR #227, opened 2024-04-07
workaround prerender in pageMap

alexanderniebuhr:
we need to plugins so we can use `generateBundle` twice. Any other hook doesn't have file names with hashes already, and using `writeBundle` would need `fs` operation, which I think is worse.

`regex` is scary, but I'm not sure if it is worth to create an ast and convert back to string again

### DIFF
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -97,7 +97,13 @@ export default function createIntegration(args?: Options): AstroIntegration {
 									// Loop through all chunks and find out which pages are prerendered
 									for (const chunk of Object.values(bundle)) {
 										if (chunk.type !== 'chunk') continue;
-										if (chunk.dynamicImports.some((entry) => entry.includes('prerender'))) {
+										if (
+											chunk.dynamicImports.some(
+												(entry) =>
+													entry.includes('prerender') &&
+													chunk.name !== '_@astrojs-ssr-virtual-entry'
+											)
+										) {
 											prerenderImports.push([chunk.facadeModuleId ?? '', chunk.fileName]);
 										}
 									}
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -94,43 +94,43 @@ export default function createIntegration(args?: Options): AstroIntegration {
 								name: 'prerender-scanner',
 								enforce: 'post',
 								generateBundle(_, bundle) {
+									// Loop through all chunks and find out which pages are prerendered
 									for (const chunk of Object.values(bundle)) {
 										if (chunk.type !== 'chunk') continue;
 										if (chunk.dynamicImports.some((entry) => entry.includes('prerender'))) {
 											prerenderImports.push(chunk.fileName);
 										}
 									}
-								},
-							},
-							{
-								name: 'prerender-worker',
-								enforce: 'post',
-								generateBundle(_, bundle) {
-									for (const chunk of Object.values(bundle)) {
-										if (chunk.type !== 'chunk') continue;
-										if (chunk.name === '_@astrojs-ssr-virtual-entry') {
-											chunk.dynamicImports = chunk.dynamicImports.filter(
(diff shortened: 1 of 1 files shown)


packages/cloudflare/src/index.ts
```diff
@@ -88,6 +90,50 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+													`^const (_page\\d) = \\(\\) => import\\('.\\/${page}'\\);$\\n`,
+													'gm'
+												);
+												let pageId: string | undefined;
+												const matches = chunk.code.matchAll(importRegex);
+												for (const match of matches) {
+													if (match[1]) {
+														pageId = match[1];
+													}
+												}
```
Fryuni, line 126: It seems this will find the last occurrence of the import. What if the same chunk is imported more than once? Is there a guarantee that it won't happen (I know we're working with very few guarantees here)

packages/cloudflare/src/index.ts
```diff
@@ -88,6 +90,50 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+												chunk.code = chunk.code.replace(importRegex, '');
+												if (pageId) {
+													const arrayRegex = new RegExp(`^^ +\\[".+", ${pageId}\\]$\\n`, 'gm');
```
Fryuni, line 129: [Regex diagram](https://regexper.com/#%2F%5E%5E%20%2B%5C%5B%22.%2B%22%2C%20_page123%5C%5D%24%5Cn%2Fgm)

image

I don't know what is the code that this Regex is trying to match, but it is looking for:
- The start of an array
- The start of a string
- Anything
- The end of a string
- The value of the variable named the value `pageId`
- The end of a line
- A following line
- Is entirely contained within a single line

And it doesn't need to match the start of the line twice (`^^` could be `^`). [Some examples](regexr.com/7ukvv).

Is that the expected pattern?

Fryuni (commented):
I don't get why this can't be done with a single plugin. Can't you do one loop after the other in the same plugin?

packages/cloudflare/src/index.ts
alexanderniebuhr: Initially I thought I needed two plugins because I first need to check all chunks in the bundle before I can mutate the entry. And the order was not right, but just looping twice inside the same plugin would work!

packages/cloudflare/src/index.ts
alexanderniebuhr: @Fryuni the goal is to match and remove from the `pageMap`, e.g.

```diff
const pageMap = new Map([
    ["../../../../../node_modules/.pnpm/astro@4.5.8_@types+node@18.17.8_typescript@5.2.2/node_modules/astro/dist/assets/endpoint/generic.js", _page0],
    ["src/pages/add/[a]/[b].ts", _page1],
-    ["src/pages/hybrid.ts", _page2]
]);
```

However we don't get `src/pages/hybrid.ts` in Vite IIUC, so the only think we know for sure to match is `_page2`. Any ideas how to either get the source filename or make the regex safer?

Fryuni: That information can be retrieved from the module graph, on `astro` it is done by [these helper functions](https://github.com/withastro/astro/blob/0fec72b35cccf80b66a85664877ca9dcc94114aa/packages/astro/src/prerender/metadata.ts).

alexanderniebuhr: @Fryuni How can I retrieve the initial source path e.g. `src/pages/hybrid.ts` inside `generateBundle`.
The helper functions astro core uses, don't work inside the adapter 🤔

alexanderniebuhr:
@Fryuni Updated the logic, now the regexp should be much safer. wdyt?

bluwy:
If this workarounds some issue with pre-rendering currently, then the code seems fine to me. But I think given the workarounds needed so far, we should prioritize making prerendering more robust in Astro core.

Fryuni (approved):

PR #231, opened 2024-04-09
fix: remove vite.ssr.noExternal because it should use default.

mrthiti:
## Changes

- Remove "vite.ssr.noExternal" because it should use default and it can append from astro config.

## Docs

Should remove this.
image

### DIFF
--- a/.changeset/rude-donkeys-fail.md
+++ b/.changeset/rude-donkeys-fail.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Remove "vite.ssr.noExternal" because it should use default and it can append from astro config.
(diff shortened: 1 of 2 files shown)


alexanderniebuhr (changes requested):
@mrthiti Thank you for the thought. But I don't think this is correct. Can you explain why you think we should remove `vite.ssr.noExternal`? This setting is needed to be compatible with Cloudflare's workerd runtime.

lilnasy:
@mrthiti I'm not sure why qrcode would break because of this setting. It's hard to say whether this is the right solution without understanding the problem first.

Could you share a minimal project that breaks with the cloudflare adapter?

mrthiti:
Hi @lilnasy
This is mini project. Could you please "npm run build", it will error.
https://github.com/mrthiti/astro-cloudflare-root-cause

alexanderniebuhr:
@mrthiti Thank you for your reproduction, I'll create a new issue for it. The solution you proposed doesn't work for Cloudflare, because Cloudflare needs to have dependencies inlined on server build, which your solution disables.

PR #232, opened 2024-04-09
fix(cloudflare): uses default values for noExternal on client

alexanderniebuhr:
## Changes

- save the default values of `noExternal` and reapply them for client

## Testing

- existing tests

## Docs

- bug fix, no docs needed

### DIFF
--- a/.changeset/rotten-ligers-rest.md
+++ b/.changeset/rotten-ligers-rest.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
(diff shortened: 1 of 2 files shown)


bluwy:
Is there a repro/example that shows this fix is required? I don't think the `ssr.*` config should have affected the client build since it's not related to SSR/server build 🤔

alexanderniebuhr:
You're right this is probably not required, but wouldn't it be safe to set it back to the initial value anyways?

bluwy:
I don't think it's needed. If the build chain at one point needs to set `noExternal: true`, then the rest should keep it the way it is.

PR #243, opened 2024-04-22
fix(cloudflare): don't cleanup client bundle

alexanderniebuhr:
## Changes

- if the plugin runs for `client` it overrides, `nonServerChunks` with an empty array, leading to not removing the unused chunks
- closes https://github.com/withastro/adapters/issues/245
- closes https://github.com/withastro/adapters/issues/244

## Testing

- I was not able to figure out why the existing tests didn't catch this, might be a super edge case 🤔 

## Docs

- internal
- bug-fix
- no docs needed

### DIFF
--- a/packages/cloudflare/src/utils/index.ts
+++ b/packages/cloudflare/src/utils/index.ts
@@ -23,7 +23,7 @@ export function mutatePageMapInPlace(
 				// @ts-expect-error - @types/estree seem to be wrong
 				s.remove(arrayExpression.start, arrayExpression.end);
 
-				// We need to check if there are any lefover commas, which are not part of the `ArrayExpression` node
+				// We need to check if there are any leftover commas, which are not part of the `ArrayExpression` node
 				// @ts-expect-error - @types/estree seem to be wrong
 				const endChar = s.slice(arrayExpression.end, arrayExpression.end + 1);
 				if (endChar.includes(',')) {
--- a/packages/cloudflare/src/utils/index.ts
+++ b/packages/cloudflare/src/utils/index.ts
@@ -22,6 +22,14 @@ export function mutatePageMapInPlace(
 			if (arrayExpression.start && arrayExpression.end) {
 				// @ts-expect-error - @types/estree seem to be wrong
 				s.remove(arrayExpression.start, arrayExpression.end);
+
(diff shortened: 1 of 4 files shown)


packages/cloudflare/src/utils/non-server-chunk-detector.ts
```diff
@@ -15,6 +15,8 @@ export class NonServerChunkDetector {
 		return {
 			name: 'non-server-chunk-detector',
 			generateBundle: (_, bundle) => {
+				// Skip if we bundle for client
+				if (!bundle['index.js']) return;
```
Fryuni, line 19: We could just not add the plugin for the client run, right?

packages/cloudflare/src/utils/index.ts
```diff
@@ -22,6 +22,14 @@ export function mutatePageMapInPlace(
 			if (arrayExpression.start && arrayExpression.end) {
 				// @ts-expect-error - @types/estree seem to be wrong
 				s.remove(arrayExpression.start, arrayExpression.end);
+
+				// We need to check if there are any lefover commas, which are not part of the `ArrayExpression` node
```
Fryuni, line 26:
```suggestion
				// We need to check if there are any leftover commas, which are not part of the `ArrayExpression` node
```

packages/cloudflare/src/utils/non-server-chunk-detector.ts
alexanderniebuhr: Well how would you do that? The client and server run use the same `vite` object 🤔

alexanderniebuhr: AFAIK  the object reference is the same. So if you add it in for the `server` target it still exists for the `client` target, and you would actually have to remove it manually again, see: https://github.com/withastro/adapters/blob/1927f941b41ac7156054c337edb031be8132e3e2/packages/cloudflare/src/index.ts#L268-L274

Fryuni: Interesting...
I remember testing something related to this and it worked. Maybe it is only a shallow clone and nested objects are shared.

Well, if it doesn' work in this case then how you did LGTM :)

Issue #254, opened 2024-05-05
can't use env in Netlify edge runtime


jamesarosen:
### Astro Info

```block
Astro                    v4.7.1
Node                     v18.19.1
System                   macOS (arm64)
Package Manager          npm
Output                   server
Adapter                  @astrojs/netlify
Integrations             none
```

### Describe the Bug

When using Netlify's Edge Runtime, Astro compiles `import.meta.env.FOO` to `process.env.FOO`, but `process.env` doesn't exist for that runtime.

#### How I found this

`astro build` adds the Astro component runtime to the site's middleware stack if both (a) the middleware and (b) a `.astro` Page or Component `import` the same shared library from `src/lib/`, even if that library has nothing to do with Astro or HTML.

In [b09bfe5a](https://github.com/jamesarosen/astro-broken-middleware-demo/commit/b09bfe5adaf79be1054030009a1b07d45e1268ba), the `middleware.ts` shares nothing with the `*.astro` files. `process.env` does not appear in the compiled `middleware.mjs`. The compiled `middleware.mjs` is 56710 bytes. It includes the `cookie` and `astro/dist/core` modules.

In [4551a260](https://github.com/jamesarosen/astro-broken-middleware-demo/commit/4551a260977f6fab6d9c9fb0825d50cb9734218e), the `middleware.ts` and `Card.astro` share an import, though the import is pure JS and unrelated to Astro. This import causes the compiled `middleware.mjs` to gain 66721 bytes, 1600 new lines, and the `cssesc`, `kleur`, `html-escaper`, and `clsx` packages. Most notably, it adds a `process.env` reference that breaks the Netlify build.

In [4fba3cf5](https://github.com/jamesarosen/astro-broken-middleware-demo/commit/4fba3cf5ea42d6c0578e914f63401e2ab34a5efd), I show that just adding an `import.meta.env.NODE_ENV` reference to the middleware is enough to break the Netlify build.

I get this in my site's logs, though the line numbers don't match up with what's compiled here:

```
7:23:50 AM: > grep -C 5 -n -e "process.env" .netlify/edge-functions/middleware/middleware.mjs
7:23:50 AM: 172-    }
7:23:50 AM: 173-  }
7:23:50 AM: 174-});
7:23:50 AM: 175-
7:23:50 AM: 176-// .netlify/functions-internal/ssr/_astro-internal_middleware.mjs
7:23:50 AM: 177:var NODE_ENV = process.env.NODE_ENV;
[...]

PR #255, opened 2024-05-06
fix(netlify): shim process.env in edge middleware

ascorbic:
## Changes

Netlify Edge Functions don't define `process.env`, but Vite uses it for any code that accesses `import.meta.env`. This PR imports the Deno `node:process` polyfill at the top of the bundled  middleware edge function

Fixes https://github.com/withastro/adapters/issues/254

## Testing

Tested with repro site https://github.com/jamesarosen/astro-broken-middleware-demo

### DIFF
--- a/.changeset/wet-dragons-count.md
+++ b/.changeset/wet-dragons-count.md
@@ -2,4 +2,4 @@
 '@astrojs/netlify': patch
 ---
 
-Define `process.env` in edge middleware
+Fixes an issue with edge middleware where `process.env` was not defined, by using a polyfill to shim it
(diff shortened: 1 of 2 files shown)


.changeset/wet-dragons-count.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Define `process.env` in edge middleware
```
alexanderniebuhr, line 5:
```suggestion
Fixes an issue with edge middleware where `process.env` was not defined, by using a polyfill to shim it
```

alexanderniebuhr (approved):
Only the changeset needs to be updated, otherwise this lgtm: )

PR #258, opened 2024-05-08
feat(cloudflare): support astro:env

alexanderniebuhr:
## Changes

- Adds support for `getSecret`

## Testing

Integration test

## Docs

N/A, I think we need to add some precisions in the config reference since it's scoped to the request


alexanderniebuhr:
cc @florian-lefebvre

florian-lefebvre:
Thanks I'll have a look!

packages/cloudflare/src/entrypoints/image-service.ts
```diff
@@ -25,6 +25,7 @@ const service: ExternalImageService = {
 		}
 
 		const imageEndpoint = joinPaths(
+			// @ts-expect-error - Property 'env' does not exist on type 'ImportMeta'.ts(2339)
```
alexanderniebuhr: I thought this was fixed with the update to the matching/newest vite version 🤔

florian-lefebvre: No this was by fixed by the `astro/client` reference

packages/cloudflare/src/utils/non-server-chunk-detector.ts
```diff
@@ -1,5 +1,5 @@
 import type { OutputBundle } from 'rollup';
-import type { Plugin } from 'vite';
+import type { Plugin, PluginOption } from 'vite';
```
alexanderniebuhr, line 2:
```suggestion
import type { PluginOption } from 'vite';
```

PR #264, opened 2024-05-17
fix(cloudflare): solid ssr

### LINKED ISSUE #224: Cloudflare v10 + Solid breaks some components and scripts
### Astro Info

```block
Astro                    v4.5.16
Node                     v20.12.0
Package Manager          bun
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             unocss
                         @astrojs/solid-js
                         @swup/astro
                         @vite-pwa/astro-integration
```

### Describe the Bug

When I deployed my web app using the new Cloudflare adapter, the functionality of many buttons and pages became broken.

For many components, it acts like I haven't added `client:load` at all - just rendering the component in a static form without having any functionality. Surprisingly, it worked for a few others. I'm still not sure about the exact situation causing it.

Also, it doesn't render the pages that have `onMount` in them. It just outputs blank.

All of them should work properly. In the minimal reproduction, you can see how using `onMount` makes the rendering blank. You can also comment `onMount` on the `Btn.tsx` component to see how it renders. I still don't have a concrete example of when the component becomes static and doesn't do anything, but I will add it here if I can.

https://stackblitz.com/~/github.com/MehrdadKhnzd/astro-bug-reproduction

### Participation

### LINKED ISSUE #263: Cloudflare + Solid-js  hydration / build errors.
### Astro Info

```block
Astro                    v4.8.4
Node                     v21.5.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/solid-js
```

### If this issue only occurs in one browser, which browser is a problem?

All

### Describe the Bug

Can't run wrangler pages dev ./dist  and get a reliable build when a component has any hydration directive that is not client:only="solid-js".    The logs say, "✘ [ERROR] Uncaught (in response) TypeError: Cannot set properties of undefined (setting 'noHydrate')". 
Min repro attached due to same error as reported by a member in discord. In my other project, I had @astrojs/solid-js 4.1.0 , @astrojs/cloudflare 10.2.4, astro 4.8.3 , solid-js 1.8.15.   I downgraded to @astrojs/cloudflare 9+, but things reliably broke on any 10+ cloudflare adapter.  

Discord link for futher viewing. 
https://discord.com/channels/830184174198718474/1239920931510554655/1239920931510554655

Any hydration directive should build and work. 

https://github.com/wkelly17/astro-min-repro

### Participation

alexanderniebuhr:
## Changes

- it seems solid & vite have an issue if bundled for webworker, added a workaround for now
- closes #224 
- closes #263

## Testing

- existing tests

## Docs

- bug fix of expected bahavior


packages/cloudflare/src/index.ts
```diff
@@ -235,6 +235,8 @@ export default function createIntegration(args?: Options): AstroIntegration {
 					// We need those conditions, previous these conditions where applied at the esbuild step which we removed
 					// https://github.com/withastro/astro/pull/7092
 					vite.resolve.conditions.push('workerd', 'worker');
+					// https://github.com/solidjs/vite-plugin-solid/issues/153
+					vite.resolve.mainFields = ['module'];
```
bluwy, line 239: I don't think we should do this. It would break many other packages too that currently relies on the default `mainFields` configuration. I think solid should be fixing this itself: https://publint.dev/solid-js@1.8.17

PR #275, opened 2024-06-06
feat(netlify): support astro:env

florian-lefebvre:
## Changes

- Adds support for Astro Env

## Testing

N/A

## Docs

N/A


packages/cloudflare/src/utils/non-server-chunk-detector.ts
```diff
@@ -1,5 +1,5 @@
 import type { OutputBundle } from 'rollup';
-import type { Plugin } from 'vite';
+import type { Plugin, PluginOption } from 'vite';
```
alexanderniebuhr, line 2:
```suggestion
import type { PluginOption } from 'vite';
```

alexanderniebuhr (approved):
Minor detail, otherwise lgtm

Issue #285, opened 2024-06-16
netlify adapter unable to handle node builtin imports


theoephraim:
### Astro Info

```block
Astro                    v4.10.2
Node                     v20.11.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/netlify
Integrations             dmno-astro-integration
```

### Describe the Bug

When using Netlify edge functions, it is possible to use node internal packages by importing them using the `node:` prefix. These imports do not work in their deno runtime otherwise.

The netlify astro adapter takes over bundling from netlify, and uses esbuild in `platfiorm: neutral` mode. In this mode, built-in node imports do not seem to supported, regardless of using the `node:` prefix or not. Here is an example, trying both a node:prefixed and unprefixed import.

<img width="749" alt="image" src="https://github.com/withastro/adapters/assets/1158956/61fceb87-e302-4e45-b628-fe9426375bf7">

note that the repro doesn't show the errors (or complete the build properly) when running on stackblitz, but if you clone the repo and run locally you'll see it.

### What's the expected result?

At the very least, I would expect that `node:` prefixed imports would work as they do on netlify. Taking it one step further, we could alter the imports of node modules that are missing the prefix to add it - which can be useful if the import was created by a dependency and not by the user themselves.

This can be easily fixed with a simple esbuild plugin
```
 await build({
  // ...
  plugins: [{
    name: 'allowNodePrefixedImports',
    setup(build) {
      build.onResolve({ filter: /^node:.*$/ }, (args) => ({ path: args.path, external: true }));
    },
  }],
  // ...
```

and an example showing a plugin that adds of the `node:` prefix. Although there could be cases where this is not the intended behaviour.
```
const NODE_BUILTIN_MODULE_REGEX = /^(assert|buffer|child_process|cluster|crypto|dgram|dns|domain|events|fs|http|https|net|os|path|punycode|querystring|readline|stream|string_decoder|timers|tls|tty|url|util|v8|vm|zlib)$/;

// plugin to add missing `node:` prefix - which is required for deno
function addNodeImportPrefix() {
  return {
    name: 'addNodeImportPrefix',
    setup(build: any) {
      build.onResolve({ filter: NODE_BUILTIN_MODULE_REGEX }, (args: any) => ({
        path: `node:${args.path}`,
        external: true,
      }))
    },
  }
}
```

### Link to Minimal Reproducible Example

[...]

PR #286, opened 2024-06-16
netlify adapter - add esbuild plugin to allow `node:` prefixed imports

theoephraim:
Addresses #285 

## Changes

Adds a simple esbuild plugin to allow `node:` prefixed imports in the netlify adapter's call to esbuild.

## Testing

Tests are currently not passing.

I started to play with it but it seems to do more with how the testing is set up and fixtures are loaded. I would have expected the middleware having node imports in it to not trip up at all, and only falter on the unprefixed node import, only when `edgeMiddleware: true` option is on.

Happy to help collaborate to fix it, but someone with more knowledge of the testing setup will be much better able to solve this.

I have personally verified that adding this code fixes my issues when building and deploying to netlify, and does not seem to break anything else.

## Docs

I think this change should match user's expectations moreso than the current behaviour.


alexanderniebuhr (commented):
The change looks good to me, we need to figure out why tests are broken. I won't have time for that soon, but I posted this in our Discord to see if anyone else wants to jump in.

alexanderniebuhr:
I will probably find time to look at the test soon. However wondering to get @ascorbic review of the general approach of allowing `node:` prefix. Should be safe, I guess :)

ascorbic:
Yes, the approach is good, and is similar to the one I used for Remix. A good future addition could potentially be to automatically add the prefix for all node builtins when reolving the ID.

alexanderniebuhr:
Looked at the testing, and I guess it's caused because we don't mark node built-ins as `external` (https://github.com/withastro/adapters/pull/286/files#diff-ee633f26523d8c890908778b42213c6fa4b86384998edff496759673f7b9ff95L312)

@ascorbic does Netlify support imports from `node_modules` folder, or do we also need to bundle all dependencies inline?

packages/netlify/test/functions/fixtures/middleware/src/middleware.ts
```diff
@@ -1,5 +1,10 @@
+import http from 'http';
```
alexanderniebuhr, line 1:
```suggestion
```

packages/netlify/test/functions/fixtures/middleware/src/middleware.ts
```diff
@@ -1,5 +1,10 @@
...
+import https from 'node:https';
+
 export const onRequest = (context, next) => {
 	context.locals.title = 'Middleware';
+	context.locals.nodeUnprefixedImportExists = !!http;
```
alexanderniebuhr, line 6:
```suggestion
```

.changeset/silly-snails-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+allow `node:` prefixed imports
```
alexanderniebuhr, line 5:
```suggestion
Allows support for `node:` prefixed imports if using  the Adapter with `edgeMiddleware: true`
```

alexanderniebuhr (approved):
LGTM

PR #290, opened 2024-06-18
Cloudflare v11

alexanderniebuhr:
## Changes

- supports Astro `>= 4.10.3`
- removes deprecated `wasmModuleImports`
- changes `platformProxy` to be enabled by default

## Testing

- existing tests

## Docs

- ⚠️ still need to check this


OliverSpeir:
What about setting [imageService](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#imageservice) to `compile` by default?

I think this would be the expected default behavior

florian-lefebvre (changes requested):
Blocking as we need to make changes for astro:env

packages/cloudflare/src/index.ts
```diff
@@ -197,7 +139,8 @@ export default function createIntegration(args?: Options): AstroIntegration {
 				});
 			},
 			'astro:server:setup': async ({ server }) => {
-				if (args?.platformProxy?.enabled === true) {
+				const usesPlatformProxy = args?.platformProxy?.enabled ?? true;
+				if (args?.platformProxy && usesPlatformProxy === true) {
```
dario-piotrowicz, line 143: is this code correct?
the `if` condition disables the use of `platformProxy` if it's not specified in the args, which I think it's not what we want?

what about something like this instead?
Screenshot 2024-06-19 at 23 44 31

packages/cloudflare/src/index.ts
```diff
@@ -225,7 +168,8 @@ export default function createIntegration(args?: Options): AstroIntegration {
 								caches: platformProxy.caches,
 								ctx: {
 									waitUntil: (promise: Promise<any>) => platformProxy.ctx.waitUntil(promise),
-									passThroughOnException: () => platformProxy.ctx.passThroughOnException(),
+									// Currently not available: https://developers.cloudflare.com/pages/platform/known-issues/#pages-functions
+									// passThroughOnException: () => platformProxy.ctx.passThroughOnException(),
```
dario-piotrowicz, line 172: We don't need to do it right away, but I am wondering, what if `platformProxy.ctx.passThroughOnException()`, when used in a Pages application, would throw an error or something warning users not to use it?

(I feel like that could be more convenient here instead of hiding the function, since I'd imagine that types would still show the function as present in any way, right?)

dario-piotrowicz (approved):
Looks great to me! 😄 

I just am not sure why you're removing the fixture files but that's just because I am lacking the context here, everything else makes sense to me 😄 

(but I also assume that the Astro docs need to be appropriately updated right?)

.changeset/fresh-walls-bow.md
```diff
@@ -0,0 +1,15 @@
+---
+'@astrojs/cloudflare': major
+---
+
+## Astro support
+
+Drops support for Astro versions `<= 4.10.2`. This allows us to remove additional workarounds and should fix all issues with bundling and prerendering. If you still observe an issue, please check current open issues or create a new one.
+
```
florian-lefebvre, line 8:
```suggestion

## Additions

Adds support for experimental `astro:env`.

```

florian-lefebvre (approved):

packages/cloudflare/src/index.ts
```diff
@@ -177,7 +184,13 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
 				}
 			},
-			'astro:build:setup': ({ vite, target }) => {
+			'astro:build:setup': async ({ vite, target }) => {
+				if (args?.platformProxy?.enabled === true) {
+					const platformProxy = await createPlatformProxy(args.platformProxy);
+
+					setProcessEnv(_config, platformProxy.env);
+				}
+
```
alexanderniebuhr, line 193: @florian-lefebvre why do we need the platformProxy inside of `astro:build:setup`?

packages/cloudflare/src/index.ts
```diff
@@ -191,16 +155,15 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							isSharpCompatible: false,
 							isSquooshCompatible: false,
 						},
+						envGetSecret: 'experimental',
 					},
 				});
 			},
 			'astro:server:setup': async ({ server }) => {
 				if (args?.platformProxy?.enabled === true) {
```
alexanderniebuhr, line 163: I need to fix this condition, to have a default value of `true`

packages/cloudflare/src/index.ts
florian-lefebvre: Yeah I took the easy way. Basically during build, it seems that `vars` in the wrangler config are not loaded so it fails although it shouldn't. We need to override `process.env` at some point in the build but yeah, idk if there's a better way to get the env there

packages/cloudflare/src/index.ts
florian-lefebvre: ah yeah my bad when I refactored I messed up that part

alexanderniebuhr:
> _https://github.com/withastro/adapters/pull/290#issuecomment-2177221153_: What about setting [imageService](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#imageservice) to compile by default?

I don't think it would be good, since it would not support image optimizations on SSR pages at all anymore.

packages/cloudflare/src/index.ts
alexanderniebuhr: @dario-piotrowicz any idea? I don't think it would be super safe to use `platformProxy` here, since you wouldn't get access to your remote configured env vars & bindings 🤔

Additionally we probably only would want to get env vars and not all bindings, but I don't think that is added yet.. I saw that enviroments support is added, correct?

OliverSpeir:
> > _[#290 (comment)](https://github.com/withastro/adapters/pull/290#issuecomment-2177221153)_: What about setting [imageService](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#imageservice) to compile by default?
> 
> I don't think it would be good, since it would not support image optimizations on SSR pages at all anymore.

But the current default is completely disabled? So that also doesn’t support SSR at all at least with compile we get as much image optimization as possible 

People will understand that they can’t run sharp in worker run time but it isn’t clear why it’s turned off when the site built with node

packages/cloudflare/src/index.ts
dario-piotrowicz: > any idea? I don't think it would be super safe to use platformProxy here, since you wouldn't get access to your remote configured env vars & bindings 🤔

sorry just so that I have the full context, what is this for? does Astro inline the env variables or something like that? (or is this about ssg?)

You're right that `platformProxy` contains local simulations so it's not really safe to use during build time, but this is true for ssg and actual bindings usage (e.g. if the code needs to be used for ssg and it can rely on a KV store), if you're only talking about env variables it can be safe-ish to use if you also enforce the use of `pages_build_output_dir` (since the local configuration will become the source of truth in that case)

> Additionally we probably only would want to get env vars and not all bindings, but I don't think that is added yet..

You're referring to https://github.com/cloudflare/workers-sdk/issues/4792 right?
If that's the case yeah it is not yet implemented... I did attempt implementing that a couple of times but I received pushbacks manly for two reasons, the fact that Astro is the only one  requesting it and the fact that we'll need to support all binding types, even future ones, and the team was concerned about the maintainability costs

I'm totally happy to revisit this, especially it if is becoming a blocker for you

> I saw that enviroments support is added, correct?

Yes, in `getPlatformProxy` you can set the environment to use (https://developers.cloudflare.com/workers/wrangler/api/#parameters-1) 🙂👍 
(but it is up to the Astro plugin to set that up)

alexanderniebuhr:
@OliverSpeir

> But the current default is completely disabled?

Not really, we default to the `noop` provider.

> People will understand that they can’t run sharp in worker runtime

That's true, but using `compile` also means that you can't use anything from Astro assets inside a SSR route, even all usage of components e.g. `<Image />` would break. I don't think many would know that. 

I think what you want is to use sharp for SSG and `noop` for SSR, which doesn't exist right now

alexanderniebuhr:
> I just am not sure why you're removing the fixture files

As a general information, the test is not working anymore in this specific case. We did add a test core, so we don't need the workaround test here anymore.

janus-reith:
I was impatient and just tried to import the v11 package into my monorepo to see if it fixes my env issues - unfortunately it didn't, the deployment still fails after the build succeeds:

```
✨ Upload complete!
--
Success: Assets published!
Error: Failed to publish your Function.
Got error: Uncaught Error: EnvInvalidVariable: The following environment variable does not match the data type and/or properties defined in `experimental.env.schema`:
MY_CONTEXT_SERVER_ACCESS_SECRET_ENV_VAR is not of type string
at chunks/server_DeNmTe1c.mjs:137:10 in createInvalidVariableError   at chunks/server_DeNmTe1c.mjs:152:8 in _internalGetSecret   at chunks/server_DeNmTe1c.mjs:163:23

```

I'm trying to deploy right from the Cloudflare Dashboard with a connected Git Repo and the mentioned env variable is present in my CF Pages Environment variables
I thought v11 would fix https://github.com/withastro/adapters/issues/294?

florian-lefebvre:
@janus-reith Let's wait for at least a pre-release

lordor5:
[ERROR] [content] [EnvInvalidVariables] The following environment variables do not match the data type and/or properties defined in `experimental.env.schema`:
Variable PUBLIC_API_URL is not of type: string.

I'm getting the same error, the workaround I found is to set optional: false in the schema.

PUBLIC_API_URL: envField.string({
          context: "client",
          access: "public",
          startsWith: "http",
          optional: false,
}),

 Is there a better fix?

florian-lefebvre:
This should not happen, can you open a new issue ?

PR #296, opened 2024-06-21
perf(netlify): handle dependency tracing for SSR function

ascorbic:
## Changes

Currently this adapter relies on the Netlify build process to bundle and trace the SSR functions. This can be slow, particularly when there are a large number of generated chunks from content collections. This PR changes the adapter to handle its own tracing using `@vercel/nft`. This is already done in the `@astrojs/vercel`, so I've mostly taken the code from there and adapted it a bit. Once we've done that we can tell Netlify to not bundle the generated function. For context, the astro.build site was taking over 6 minutes to bundle before this. With this PR, the bundling takes 6 seconds

See https://app.netlify.com/teams/ascorbic/builds/667d304a908d450008617d30 for an example build


ascorbic:
!preview

ascorbic:
!preview netlify-nft

ascorbic:
!preview netlify-nft

ascorbic:
!preview netlify-nft

packages/netlify/src/lib/nft.ts
```diff
@@ -0,0 +1,85 @@
...
+			// Sharp is always external and won't be able to be resolved, but that's also not a problem
+			if (module === 'sharp') continue;
+
+			if (entryPath === file) {
+				logger.debug(
+					`[@astrojs/netlify] The module "${module}" couldn't be resolved. This may not be a problem, but it's worth checking.`
+				);
+			} else {
+				logger.debug(
+					`[@astrojs/netlify] The module "${module}" inside the file "${file}" couldn't be resolved. This may not be a problem, but it's worth checking.`
```
ematipico, line 63: `[@astrojs/netlify]` isn't needed. The logger provided to the adapter is already scoped to prefix `@astrojsd/netlify` to each log

packages/netlify/src/index.ts
```diff
@@ -332,6 +363,8 @@ export default function netlifyIntegration(
 			get cookies(): never {
 				throw new Error('Please use Astro.cookies instead.');
 			},
+			// @ts-expect-error
```
ematipico, line 366: nit: it could be useful to add a comment that explains why we should except an error here

ematipico (approved):

ascorbic:
WINDOWS PATHS 😱

PR #315, opened 2024-07-01
feat: use Netlify Frameworks API

eduardoboucas:
## Changes

Updates the Netlify adapter to use the upcoming Frameworks API.

## Testing

Tests updated.

## Docs

Internal change, no user-facing changes are expected.

### DIFF
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -2,7 +2,6 @@ import { randomUUID } from 'node:crypto';
 import { appendFile, mkdir, readFile, rm, writeFile } from 'node:fs/promises';
 import type { IncomingMessage } from 'node:http';
 import { fileURLToPath } from 'node:url';
-import { inspect } from 'node:util';
 import { emptyDir } from '@astrojs/internal-helpers/fs';
 import { createRedirectsFromAstroRoutes } from '@astrojs/underscore-redirects';
 import type { Context } from '@netlify/functions';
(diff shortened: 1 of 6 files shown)


packages/netlify/src/index.ts
```diff
@@ -250,20 +259,9 @@ export default function netlifyIntegration(
 					cacheOnDemandPages: Boolean(integrationConfig?.cacheOnDemandPages),
 					notFoundContent,
 				})});
-				export const config = { name: "Astro SSR", generator: "@astrojs/netlify@${packageVersion}", path: "/*", preferStatic: true };
+				export const config = ${inspect(config)};
```
eduardoboucas, line 262: I've used `inspect` and not `JSON.stringify` because we had an issue with accept string literals as keys. This was fixed in https://github.com/netlify/build/pull/5745, but it's not something we can back-port to older versions of the Netlify CLI.

packages/netlify/src/index.ts
```diff
@@ -241,6 +242,14 @@ export default function netlifyIntegration(
 			},
 			TRACE_CACHE
 		);
+		const config = {
```
matthewp, line 245: it doesn't look like this is ever used

ascorbic: It's used on line 262

packages/netlify/src/index.ts
ascorbic: Onlythe version is dynamic, so would it be cleaner to inline the whole thing like before?

eduardoboucas: Done in https://github.com/withastro/adapters/pull/315/commits/b8186e7b25507770749fb82f505b41f563a3e946.

ascorbic (approved):

PR #325, opened 2024-07-11
feat(netlify): support included/excluded files

ascorbic:
## Changes

Adds "includedFiles" and "excludedFiles" options. These allow extra files to be deployed in the SSR function bundle. It globs these files at build time, and includes them in the file list that is deployed in the function. This PR also ensures that process.cwd matches the site root during SSR.

## Testing

Adds new test fixture and tests.

## Docs

This needs docs. The changeset includes a lot of the required info, as do the types.


ascorbic:
!preview excluded-files

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+Adds "includedFiles" and "excludedFiles" options. These allow extra files to be deployed in the SSR function bundle.
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary. The `"includedFiles"` and `"excludedFiles"` options allow you to do these. Both options support glob patterns, so you can include/exclude multiple files at once.
+
+The paths are relative to the site root. If you are loading them using filesystem funcitons, make sure you resolve paths relative to the site root, and not relative to the cxource file. At runtime the compiled file will be in a different location, so paths that are relative to the file will not work. You should instead resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root.
```
florian-lefebvre, line 9: I think it's worth showing a little snippet of do/don't

packages/netlify/src/index.ts
```diff
@@ -173,6 +173,18 @@ export interface NetlifyIntegrationConfig {
...
+	 * Extra files to include in the SSR function bundle. Paths are relative to the project root.
+	 * Glob patterns are supported.
+	 */
+	includedFiles?: string[];
+
+	/**
+	 * Files to exclude from the SSR function bundle. Paths are relative to the project root.
+	 * Glob patterns are supported.
+	 */
+	excludedFiles?: string[];
```
florian-lefebvre, line 187: What was requested by the user is a way to include files right? Why would someone one to exclude some files if it's already the case by default?

packages/netlify/src/ssr-function.ts
```diff
@@ -24,7 +25,13 @@ export const createExports = (manifest: SSRManifest, { middlewareSecret }: Args)
 		cacheOnDemandPages: boolean;
 		notFoundContent?: string;
 	}) {
+		// The entrypoint is created in `.netlify/v1/build` so we need to go up two levels to get to the root
+		const root = fileURLToPath(new URL('../../', import.meta.url));
 		return async function handler(request: Request, context: Context) {
+			// This entrypoint will be deep inside the directory structure below cwd
+			// We chdir to the root so that we can resolve the correct paths at runtime
+			process.chdir(root);
```
florian-lefebvre, line 33: Are we sure it can't break any existing app?

ascorbic: As far as I can work out, existing apps that rely on cwd would either have unchanged behaviour (if they're not a monorepo), or would already be broken (if they are using a monorepo)

packages/netlify/src/index.ts
ascorbic: Two cases: either they've added a glob but want to exclude one specific file from that. This is a more intuitive way than using a negated glob. The other is to exlude automatically traced dependencies, like binaries in node moduels that it incorrectly thinks are needed. This could be used to exclude them.

.changeset/itchy-donuts-love.md
ascorbic: Good idea

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,21 @@
+---
+'@astrojs/netlify': minor
+---
+
+Adds "includedFiles" and "excludedFiles" options. These allow extra files to be deployed in the SSR function bundle.
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary. The `"includedFiles"` and `"excludedFiles"` options allow you to do these. Both options support glob patterns, so you can include/exclude multiple files at once.
```
sarah11918, line 7:
```suggestion
When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.

The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the project root. Both options support glob patterns, so you can include/exclude multiple files at once.
```
Just thought this paragraph was getting a *little* long, so maybe breaking it up to separate where the use case is described vs the options themselves?

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,21 @@
...
+
+```js
+import netlify from '@astrojs/netlify';
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  output: 'server',
+  adapter: netlify({
+    includedFiles: ['files/**/*.csv', 'files/include-this.txt'],
+    excludedFiles: ['files/subdirectory/not-this.csv'],
```
sarah11918, line 19: Not blocking, but just throwing it out there: since most people's projects are maybe not likely to have a folder actually called `files/`, I'm whether there's a starting directory that represents something someone *would* have in their project, and these inclusions/exclusions could be shown for a specific reason?

(I snagged from the code to know that these are relative to the project root, because I wasn't immediately sure what or where `files/` might be, since it's not a common example we use! But, with that added, the example is probably easy enough to grok and I'll let you decide.)

For example, I think it *could* be an idea to show excluding something from node modules, if that's something people are likely to use this for.

The `/subdirectory/` and `include-this`/`not-this` style names are great! (Even something bordering on the style of `not-this-way-huge-binary` e.g., if you're feeling cheeky!)

sarah11918 (commented):
This looks super helpful, @ascorbic and the docs are fabulous! 

Just a couple of quick thoughts mostly on formatting. 

(In particular, we do tend to use inline code when naming a feature, not quotation marks. And both inline AND quotation marks are reserved for when that's how the feature would be shown in the reader's code (e.g. `output: 'server'`. (Though, we're more relaxed with `hybrid` and `server` just written on their own.)

Otherwise, great intro, great explanation, and helpful code example. What more could I possibly want? :smile:

.changeset/itchy-donuts-love.md
ascorbic: Great suggestions! I changed it to:

```js
    includedFiles: ['src/address-data/**/*.csv', 'src/include-this.txt'],
    excludedFiles: ['node_modules/chonky-module/not-this-massive-file.mp4']
```

ascorbic:
!preview include-files

alexanderniebuhr:
@ascorbic Do you want to land this before the next minor of the Netlify Adapter?

ascorbic:
@alexanderniebuhr Ideally, yes. It's still waiting on an approval

alexanderniebuhr:
I already looked at it and quite liked it, but I can review it later today, so we can land it before we cut the release :)

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,23 @@
...
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
+
+The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the project root. Both options support glob patterns, so you can include/exclude multiple files at once.
+
+The paths are relative to the site root. If you are loading them using filesystem funcitons, make sure you resolve paths relative to the site root, and not relative to the cxource file. At runtime the compiled file will be in a different location, so paths that are relative to the file will not work. You should instead resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root.
```
bluwy, line 11:
```suggestion
The paths are relative to the site root. If you are loading them using filesystem functions, make sure you resolve paths relative to the site root, and not relative to the source file. At runtime the compiled file will be in a different location, so paths that are relative to the file will not work. You should instead resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root.
```

packages/netlify/src/index.ts
```diff
@@ -237,18 +249,37 @@ export default function netlifyIntegration(
...
+	async function getFilesByGlob(
+		include: Array<string> = [],
+		exclude: Array<string> = []
+	): Promise<Array<URL>> {
+		const files = await glob(include, {
+			cwd: fileURLToPath(rootDir),
+			absolute: true,
+			ignore: exclude,
+		});
+		return files.map((file) => pathToFileURL(file));
```
ematipico, line 261: It may be an issue. But I think if `include` is empty, then no glob is executed, which means that the `exclude` won't be picked up. I tried to look in the `fast-glob` docs, but I couldn't find anything.

packages/netlify/test/functions/fixtures/includes/astro.config.mjs
```diff
@@ -0,0 +1,11 @@
+import netlify from '@astrojs/netlify';
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  output: 'server',
+  adapter: netlify({
+    includedFiles: ['files/**/*.csv', 'files/include-this.txt'],
+    excludedFiles: ['files/subdirectory/not-this.csv'],
```
ematipico, line 8: Can we create tests cases where only `includedFiles` and `excludedFiles` are provided? E.g.

```js
adapter: netlify({
	includedFiles: [""]
})


adapter: netlify({
	excludedFiles: [""]
})
```

packages/netlify/src/index.ts
ascorbic: It's handled [here](https://github.com/withastro/adapters/blob/include-files/packages/netlify/src/index.ts#L275-L286), where the excluded files are globbed and passed to `copyDependenciesToFunction`, which skips any that match. This means it excludes files that are automatically traced as well as ones included manually.

packages/netlify/src/ssr-function.ts
ascorbic: If a site uses filesystem functions at runtime, this is the only way it can correctly load the files

packages/netlify/test/functions/fixtures/includes/astro.config.mjs
ascorbic: Added a new test fixture to show excluding traced node modules

packages/netlify/src/ssr-function.ts
ematipico, line 33: `new URL` and `process.chdir` are likely to fail, I would play it a bit more conservative: use `URL.canParse` to check if it's a valid a URL, and a `try/catch` for `process.chdir` to at least not fail (or fail) and gracefully log the error. Even Node.js docs recommend using the `try/catch` block https://nodejs.org/api/process.html#processchdirdirectory

ascorbic: What are you suggesting we check with `URL.canParse`? `import.meta.url`?

ematipico: The whole thing I suppose: 

```js
if (URL.canParse("../../", import.meta.url)) {}
```

ascorbic: What would be the scenario where it wouldn't work though, when `"../.."` is a static value, and `import.meta.url` comes from the runtime?

ematipico: There was a bug in the Cloudflare adapter where `import.meta.url` was throwing an error because it wasn't a URL. I am not saying that Netlify will get wrong, just playing it safe 😅

bluwy: I don't think using filesystem functions is something Astro supports in the runtime? Files could be moved during bundling and it wouldn't work later.

ascorbic: If `import.meta.url` is broken I don't think we should hide that!

ascorbic: I've added a try/catch, but don't think we need the URL check

ascorbic: We don't prevent people using fs functions, so I'd imagine lots of people do. This PR was prompted by an issue where a library was trying to load i18n data and it wasn't found. This part fixes precisely that issue: of files being moved during bundling. Without that there's not a lot of point to allowing users to include files in the bundle, because the user code won't be able to find them.

alexanderniebuhr (approved):
I miss specific platform knowledge but I like the feature and implementation looks good and has tests.

packages/netlify/src/ssr-function.ts
bluwy: I don't feel like fs functions are a common thing especially when dealing in a bundled runtime. There's Vite APIs like `import.meta.glob`, `?url`, and `?raw` for them. Fs functions are also likely to fail elsewhere than only on Netlify, and so we shouldn't support and encourage them.

I figured the benefit with `includeFiles` and `excludeFiles` are like Vercel's option, because nft has trouble sometimes detecting every file.

I'm not really comfortable with keeping this in the adapter. It feels surprising to me that it could change the cwd internally and on every request.

alexanderniebuhr: I think the issue with the Cloudflare adapter was not that `import.meta.url` wasn't an URL, I think it is just not supported in Cloudflare's runtime, therefore not working for `new URL()`

ascorbic: They _are_ like Vercel's option. In fact they use almost the same implementation. With libraries that use fs functions, they usually accept a directory as an option or resolve relative to cwd. It's very common for i18n libraries such as i18next, for example.

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,23 @@
+---
+'@astrojs/netlify': minor
+---
+
+Adds "includedFiles" and "excludedFiles" options. These allow extra files to be deployed in the SSR function bundle.
```
sarah11918, line 5:
```suggestion
Adds `includedFiles` and `excludedFiles` configuration options. These allow extra files to be deployed in the SSR function bundle.
```

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,23 @@
...
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
```
sarah11918, line 7:
```suggestion
When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However, sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
```

.changeset/itchy-donuts-love.md
```diff
@@ -0,0 +1,23 @@
...
+
+The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the project root. Both options support glob patterns, so you can include/exclude multiple files at once.
+
+The paths are relative to the site root. If you are loading them using filesystem functions, make sure you resolve paths relative to the site root, and not relative to the source file. At runtime the compiled file will be in a different location, so paths that are relative to the file will not work. You should instead resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root.
```
sarah11918, line 11:
```suggestion
The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the site root. Both options support glob patterns, so you can include/exclude multiple files at once.

If you are specifying files using filesystem functions, resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root. At runtime, compiled source files will be in a different location and you cannot rely relative file paths.
```

Just noting the above seemed a little repetative (array of paths relative to the site root.. the paths are relative to the site root... resolve paths relative to the site root...) :smile: 

Something like the above maybe conveys the info more compactly, and doesn't make people read more than necessary to get the info they need? (Will note that as a *changeset*, this doesn't have to be full, cover-everything, documentation either! We can spell things out more clearly and exhuastively in docs itself if you think that's necessary!)

Do you think something like this does a good enough "changelog" job?

sarah11918 (commented):
Just left a suggestion of a more compact style, if you'd like one!

alexanderniebuhr:
I was not able to wait for this PR to make it into the release, I needed to ship Cloudflare bugfixes.. Are all discussions resolved?

ascorbic:
!preview include-files

ascorbic:
!preview include-files

packages/netlify/src/ssr-function.ts
ascorbic: If we're about to do a major I can do a breaking changeset for the chdir part.

PR #326, opened 2024-07-11
feat(netlify): set cache headers for assets

ascorbic:
## Changes

Adds immutable cache headers for everything in the assets directory. This is safe because they all have hashed filenames

## Testing

Added a test, and manually checked a deployed site

### DIFF
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -112,6 +112,17 @@ async function writeNetlifyFrameworkConfig(config: AstroConfig, logger: AstroInt
 			.filter(Boolean as unknown as (pattern?: string) => pattern is string)
 	);
 
+	const headers = config.build.assetsPrefix
+		? undefined
+		: [
+				{
+					for: `${config.base}${config.base.endsWith('/') ? '' : '/'}${config.build.assets}/*`,
+					values: {
+						'Cache-Control': 'public, max-age=31536000, immutable',
+					},
+				},
+			];
+
 	// See https://docs.netlify.com/image-cdn/create-integration/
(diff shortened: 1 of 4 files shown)


packages/netlify/src/index.ts
```diff
@@ -112,13 +112,25 @@ async function writeNetlifyFrameworkConfig(config: AstroConfig, logger: AstroInt
 			.filter(Boolean as unknown as (pattern?: string) => pattern is string)
 	);
 
+	const headers = config.build.assetsPrefix
+		? undefined
+		: [
+				{
+					for: `${config.base}${config.base.endsWith('/') ? '' : '/'}${config.build.assets}/*`,
```
florian-lefebvre, line 119: What happens if base isn't set? Also maybe worth extracting to withTrailingSlash if such logic is used elsewhere

packages/netlify/src/index.ts
```diff
@@ -432,9 +444,7 @@ export default function netlifyIntegration(
 				rootDir = config.root;
 				_config = config;
 
-				if (config.image?.domains?.length || config.image?.remotePatterns?.length) {
-					await writeNetlifyFrameworkConfig(config, logger);
-				}
+				await writeNetlifyFrameworkConfig(config, logger);
```
alexanderniebuhr, line 447: Is it safe to remove the if condition?

alexanderniebuhr (approved):
One question, otherwise lgtm

packages/netlify/src/index.ts
ascorbic: Yeah, it was only there because previously it was only needed for images. As we're now using it for headers we need to write it every time

matthewp (approved):

Issue #266, opened 2024-05-20
`compiled` image option should create passthrough endpoint for built files


bholmesdev:
### Astro Info

```block
Astro                    v4.8.4
Node                     v20.13.1
System                   macOS (arm64)
Package Manager          pnpm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/markdoc
                         astro-icon
```

### Describe the Bug

I am building a blog with Cloudflare workers and Astro content collections. Migrating from a static setup to SSR, I wanted to preserve image optimization. Content collections are processed at build time so I can luckily use Sharp during the build to avoid issues in SSR. However, [Cloudflare's `compiled` image service](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#imageservice) doesn't quite support this pattern at runtime.

My setup:
- I have a content collection called `blog` containing my posts and related image assets
- I have a server-rendered route (**not prerendered**) that fetches each post with `getEntry()`

The build succeeds running sharp locally as expected. However, Astro generates image path to resolves to an SSR endpoint (`<img src="/_image?href=/_astro/image.png"`), rather than the built path (`<img src="/_astro/image.png"`).

### What's the expected result?

I expect Astro to output this endpoint route (`/_image`) since I am not using the `prerender` flag. However, it would be nice of the Cloudflare integration added a proxy endpoint that routed request to `_image` to the appropriate entry in `dist/`. I added this proxy to my own setup [using the `image.endpoint` option](https://docs.astro.build/en/reference/configuration-reference/#imageendpoint) to resolve the issue. Here is the implementation:

```ts
// src/pages/image-passthrough.ts
import type { APIRoute } from "astro";

export const prerender = false;

export const GET: APIRoute = (ctx) => {
  const href = ctx.url.searchParams.get("href");
  if (!href) {
    return new Response("Missing 'href' query parameter", { status: 400 });
  }

  if (import.meta.env.PROD && !href.startsWith("/_astro")) {
    return new Response("Invalid 'href' query parameter", { status: 400 });
  }

  return fetch(new URL(href, ctx.url.origin));
};
```

Injecting this would make `imageServer: 'compiled'` the perfect option for content collection users, since it would unlock SSR rendering as well.

Note: [I attempted to make a reproduction in Stackblitz](https://stackblitz.com/~/edit/github-vn1eik?file=astro.config.mjs)... but I couldn't get a project to build using the cloudflare adapter. I just see a hang and no dist output 😬

[...]

Issue #213, opened 2024-03-29
Cloudflare v10 breaks cloudflare build / wrangler preview `imageService: "compile", and `<Image />` component


dallyh:
### Astro Info

```block
Astro                    v4.5.12
Node                     v20.12.0
System                   Windows (x64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/react
```


### Describe the Bug

When using v10 version of the adapter, hybrid output mode and `<Image/>` component on a prerendered path, with combination of `imageService: "compile"` in the configuration, wrangler preview fails. This was not the case with previous version, it looks like there are modules being bundled which should not be bundler.

Steps to reproduce
- npm run build-and-preview

```
Attaching additional modules:
┌──────────────────────────────────────────┬──────┬───────────┐
│ Name                                     │ Type │ Size      │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/astro/assets-service_Dj66YdWb.mjs │ esm  │ 12.29 KiB │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/astro_DBzb2nRr.mjs                │ esm  │ 83.90 KiB │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/image-endpoint_RvDj3-CK.mjs       │ esm  │ 0.18 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/index_B6euTALA.mjs                │ esm  │ 0.18 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/pages/image-endpoint_nxDOIah1.mjs │ esm  │ 0.06 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/pages/index_CPLiUgeG.mjs          │ esm  │ 0.20 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/prerender_H0Lv92GL.mjs            │ esm  │ 0.09 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/test_2VqMiBwm.mjs                 │ esm  │ 0.17 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
│ chunks/vnode-children_VTcTc4QR.mjs       │ esm  │ 2.94 KiB  │
├──────────────────────────────────────────┼──────┼───────────┤
[...]

PR #335, opened 2024-07-23
fix(cloudflare): add proxy /_image endpoint

alexanderniebuhr:
## Changes

- The build succeeds running sharp locally as expected. However, Astro generates image path to resolves to an SSR endpoint (<img src="/_image?href=/_astro/image.png"), rather than the built path (<img src="/_astro/image.png")
- However, it would be nice of the Cloudflare integration added a proxy endpoint that routed request to _image to the appropriate entry in dist/
- closes https://github.com/withastro/adapters/issues/266
- maybe https://github.com/withastro/adapters/issues/213

## Testing

- added new test

## Docs

- fix, not needed


packages/cloudflare/src/entrypoints/image-endpoint.ts
```diff
@@ -1,2 +1,16 @@
...
+
+export const prerender = false;
+
+export const GET: APIRoute = (ctx) => {
+	const href = ctx.url.searchParams.get('href');
+	if (!href) {
+		return new Response("Missing 'href' query parameter", { status: 400 });
+	}
+
+	if (!href.startsWith('/_astro')) {
```
Princesseuh, line 11: You might not want to hardcode this, as it can theoretically be different based on user config

packages/cloudflare/src/entrypoints/image-endpoint.ts
```diff
@@ -1,2 +1,16 @@
...
-// it allows use to offer `imageService: 'compile'`
+import type { APIRoute } from 'astro';
+
+export const prerender = false;
+
+export const GET: APIRoute = (ctx) => {
+	const href = ctx.url.searchParams.get('href');
+	if (!href) {
+		return new Response("Missing 'href' query parameter", { status: 400 });
+	}
```
ematipico, line 9:
```suggestion
	if (!href) {
		return new Response("Missing 'href' query parameter", { status: 400, statusText: "Missing 'href' query parameter" });
	}
```

packages/cloudflare/src/entrypoints/image-endpoint.ts
```diff
@@ -1,2 +1,16 @@
...
+
+	if (!href.startsWith('/_astro')) {
+		return new Response("Invalid 'href' query parameter", { status: 400 });
```
ematipico, line 12: I would emit a different error from the previous, otherwise debugging becomes difficult, and we don't tell the user **why** the `href` is invalid.

PR #340, opened 2024-07-30
 fix(cloudflare): env vars in middleware & prerendering

alexanderniebuhr:
## Changes

- closes https://github.com/withastro/adapters/issues/337

## Testing

- existing tests should work

## Docs

- bugfix not needed


packages/cloudflare/src/index.ts
```diff
@@ -123,10 +131,15 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							cloudflareModulePlugin,
 						],
 					},
+					integrations: [astroWhen()],
```
Fryuni, line 134: It should check if the integration is already in the project to avoid duplicates. It would still work but re-do work unnecessarily

packages/cloudflare/src/entrypoints/middleware.ts
```diff
@@ -0,0 +1,15 @@
...
+		if (ctx.locals.runtime === undefined) {
+			ctx.locals.runtime = {
+				env: process.env,
+			};
+		}
+		return next();
+	},
+};
+
+export const onRequest = middlewares[whenAmI];
```
Fryuni, line 15: I _think_ Astro will fail to load an undefined injected middleware.

```suggestion
export const onRequest = middlewares[whenAmI] ?? ((_, next) => next());
```

I haven't tried it, if it works just ignore this.

Fryuni (approved):

matthewp (approved):

packages/cloudflare/src/index.ts
alexanderniebuhr: Not blocking, once I understand the issue and the possible solution, I'll can sent a follow-up PR.
However this text from the API reference, let me think, that `updateConfig` does not add integrations twice 🤔

> Any config you provide will be merged with the user config + other integration config updates.

PR #350, opened 2024-08-15
Apply polyfills right away

matthewp:
## Changes

- Same change we made to the Node adapter earlier.
- Prevents race condition where `crypto` could be used before the polyfill is applied.

## Testing

- Tested via the server-islands project.

## Docs

N/A, bug fix


packages/netlify/src/index.ts
alexanderniebuhr, line 1: Those changes are unrelated, but probably still valid. It's just hard to tell if they break anything, but tests seem fine.

alexanderniebuhr (approved):
lgtm

packages/netlify/src/index.ts
matthewp: They are unused imports/variables, I just caught them in my editor and thought why not fix.

PR #367, opened 2024-08-29
chore: move `vercel` & `node` to next branch

alexanderniebuhr:
## Changes

- https://github.com/withastro/astro/pull/11866


ematipico (changes requested):
Blocking to prevent accidental merges. cc @matthewp this PR should be merged **without** Squash, so we can keep the commit history

PR #373, opened 2024-09-02
fix: ISR opt out for routes with rest parameters

hrabiel:
## Changes

The logic for generating a match pattern ([getMatchPattern](https://github.com/withastro/adapters/blob/8274001d0d1a857e633f456c5f0ff1d469d4b6e4/packages/vercel/src/lib/redirects.ts#L46)) was initially copied from the core astro repository. Since then, the implementation in Astro Core was changed a couple of times. Here is the current implementation of this function in Astro Core: https://github.com/withastro/astro/blob/cd54210/packages/astro/src/core/routing/manifest/pattern.ts#L3.

Looking into `.vercel/output/config.json`, it was found that the route generated for the excluded path contains an extra slash. Here is an example for a `"/excluded/[...rest]"` route:
```json
...
{
	"src": "^/excluded/(?:\\/(.*?))?$",
	"dest": "_render"
},
...
```
However, it should look like this:
```json
...
{
	"src": "^/excluded(?:\\/(.*?))?$",
	"dest": "_render"
},
...
```

- Changed `getMatchPattern()` so that it doesn't put a redundant slash before a route segment with rest/spread parameters.
- Changed `getMatchPattern()` so that it also works for route segments that includes the rest parameter **not** in the beginning of the segment (e.g. `/one/two[...three]`).

## Testing

Was tested by:
- Adding a modified version of the `@astro/vercel` package to the example project.
- Deploying it to Vercel and making sure that all excluded routes are not cached.
- Checking a `.vercel/output/config.json` an making sure that the route with the rest parameter is added with the `"_render"` dest and the `"src"` regex is no longer contains an extra prepending slash.

## Docs

No need to change the docs because it already mentions this case:
docs-excluding-paths-from-caching

### DIFF
--- a/packages/vercel/src/lib/redirects.ts
+++ b/packages/vercel/src/lib/redirects.ts
@@ -45,25 +45,28 @@ function getParts(part: string, file: string) {
 // 2022-04-26
 function getMatchPattern(segments: RoutePart[][]) {
 	return segments
-		.map((segment) => {
-			return segment[0].spread
+		.map((segment, segmentIndex) => {
+			return segment.length === 1 && segment[0].spread
 				? '(?:\\/(.*?))?'
-				: segment
+				: (segmentIndex === 0 ? '' : '/') +
+					segment
 						.map((part) => {
 							if (part)
-								return part.dynamic
-									? '([^/]+?)'
-									: part.content
-											.normalize()
-											.replace(/\?/g, '%3F')
-											.replace(/#/g, '%23')
-											.replace(/%5B/g, '[')
-											.replace(/%5D/g, ']')
-											.replace(/[*+?^${}()|[\]\\]/g, '\\$&');
+								return part.spread
+									? '(.*?)'
+									: part.dynamic
+										? '([^/]+?)'
+										: part.content
+												.normalize()
+												.replace(/\?/g, '%3F')
+												.replace(/#/g, '%23')
+												.replace(/%5B/g, '[')
+												.replace(/%5D/g, ']')
+												.replace(/[*+?^${}()|[\]\\]/g, '\\$&');
 						})
 						.join('');
 		})
-		.join('/');
+		.join('');
 }
 
 function getReplacePattern(segments: RoutePart[][]) {
--- a/packages/vercel/src/lib/redirects.ts
+++ b/packages/vercel/src/lib/redirects.ts
@@ -48,10 +48,10 @@ function getMatchPattern(segments: RoutePart[][]) {
 		.map((segment, segmentIndex) => {
 			return segment.length === 1 && segment[0].spread
 				? '(?:\\/(.*?))?'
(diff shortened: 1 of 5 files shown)


matthewp (approved):

Princesseuh (approved):

packages/vercel/src/lib/redirects.ts
```diff
@@ -45,25 +45,28 @@ function getParts(part: string, file: string) {
...
-									: part.content
-											.normalize()
-											.replace(/\?/g, '%3F')
-											.replace(/#/g, '%23')
-											.replace(/%5B/g, '[')
-											.replace(/%5D/g, ']')
-											.replace(/[*+?^${}()|[\]\\]/g, '\\$&');
-						})
-						.join('');
+				: (segmentIndex === 0 ? '' : '/') +
```
bholmesdev, line 51: I noticed this check differs from the core implementation. I believe you said this address the case where a spread param lacks a leading slash (e.g. `/path[...spread]`). Is there a reason Vercel needs special handling?

hrabiel: @bholmesdev great question!
Not exactly. This a replacement for `join('/')` defined here: https://github.com/withastro/adapters/blob/2246e6e263ba5d4c67912894c011dc64cd906634/packages/vercel/src/lib/redirects.ts#L66
It means "join all route segments with `'/'`, but omit it if the segment is a spread" (e.g. "/path/**[...spread]**").

And the case you are describing is handled here (I should probably cover it with tests as well): https://github.com/withastro/adapters/blob/2625aa3d049b3c883b07017e29bace4baff979f5/packages/vercel/src/lib/redirects.ts#L56

Initially I copied the core implementation of `getPattern()` method and tried to use that, but I realized that it would require more code changes around redirects. I've tried to achieve the goal without changing the interface of `getMatchPattern()` function. If you think that the correct way is to copy `getPattern()` from core and adjust the rest of the Vercel adapter code accordingly, let me know!

bholmesdev: Got it. I can see why we took this route to avoid a larger refactor. I think that's the right move for this PR! I'd consider exposing a utility for this from `astro` core to consolidate code in the future.

bholmesdev, line 51:
```suggestion
                                 // Omit leading slash if segment is a spread.
                                 // This is handled using a regex in Astro core.
                                 // To avoid complex data massaging, we handle in-place here.
				: (segmentIndex === 0 ? '' : '/') +
```

bgentry:
FYI, it appears that this PR has [caused](https://github.com/withastro/adapters/issues/418#issuecomment-2567258474) at least some of the redirect issues reported in #418. Reverting just this commit on the latest release fixes that issue for me, though I'm not sure what specific part of the diff causes the problem.

PR #385, opened 2024-09-10
chore(next): astro:env cleanup

florian-lefebvre:
## Changes

- Cleans some forgotten workarounds for astro:env

## Testing

N/A

## Docs

N/A


packages/netlify/src/ssr-function.ts
```diff
@@ -2,14 +2,10 @@ import type { Context } from '@netlify/functions';
 import type { SSRManifest } from 'astro';
 import { App } from 'astro/app';
 import { applyPolyfills } from 'astro/app/node';
+import { setGetEnv } from 'astro/env/setup';
```
matthewp, line 5: Polyfills need to be a applied before `astro/env/setup` is loaded due to this loading code that depends on `crypto`.

What i would do is, move the `applyPolyfills()` call to another module, say `polyfill.ts` like this:

```js
import { applyPolyfills } from 'astro/app/node';

applyPolyfills();
```

Then in this file you can just import that before astro/env:

```js
import './polyfill';
import { setGetEnv } from 'astro/env/setup';
```

matthewp (changes requested):
Make sure the polyfills are applied before astro/env/setup is loaded.

PR #387, opened 2024-09-11
Added windows support for static files

veitbjarsch:
## Changes
- Added windows support for static files

On windows machines it did not split up the path for static files into it's parts, because it was only searching for posix separators.

## Testing
Run the build on windows machines an check the _routes.json before and after.

### DIFF
--- a/packages/cloudflare/src/utils/generate-routes-json.ts
+++ b/packages/cloudflare/src/utils/generate-routes-json.ts
@@ -2,7 +2,7 @@ import type { AstroConfig, AstroIntegrationLogger, RouteData, RoutePart } from '
 
 import { existsSync } from 'node:fs';
 import { writeFile } from 'node:fs/promises';
-import { posix, win32 } from 'node:path';
+import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import {
 	prependForwardSlash,
@@ -186,7 +186,7 @@ export async function createRoutesFile(
 			const staticPath = staticFile;
 
 			const segments = removeLeadingForwardSlash(staticPath)
-				.split(process.platform === 'win32' ? win32.sep : posix.sep)
+				.split(path.sep)
 				.filter(Boolean)
 				.map((s: string) => {
 					return getParts(s);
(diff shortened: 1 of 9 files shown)


veitbjarsch:
@alexanderniebuhr fixed the formatting and linting error. PR should be good to go now.

bluwy:
Do you have a repro or test for this? Checking the Astro source code, I don't think a backslash could happen in the first place. It'd be a bug there.

veitbjarsch:
@bluwy Here is a minimal reproduction of the issue. Just installed asto (latest version) and the astro cloudflare plugin.
https://github.com/veitbjarsch/astro-cloudflare-demo

When I add folders in the public folder and run `npx astro build` on windows the reoutes.json looks like this:
https://github.com/veitbjarsch/astro-cloudflare-demo/blob/main/dist/_routes.json

bluwy:
With your adjusted code, it makes more sense now and I think that's the right fix. Initially you made the change for Astro's `paths` which shouldn't have this issue.

packages/cloudflare/src/utils/generate-routes-json.ts
```diff
@@ -186,7 +186,7 @@ export async function createRoutesFile(
 			const staticPath = staticFile;
 
 			const segments = removeLeadingForwardSlash(staticPath)
-				.split(posix.sep)
+				.split(process.platform === 'win32' ? win32.sep : posix.sep)
```
bluwy, line 189: Could we simplify this as:

```js
import path from 'node:path'

//...
.split(path.sep)
```

veitbjarsch: Yes looks like this will work too. I'll adjust it on Monday.

veitbjarsch: @alexanderniebuhr @bluwy done.
Can you run the workflow tests once more.

bluwy (approved):

alexanderniebuhr:
@veitbjarsch can you add an changeset with `pnpm exec changeset`

veitbjarsch:
@alexanderniebuhr done

alexanderniebuhr:
@veitbjarsch I can't see it, have you pushed your commit?

PR #388, opened 2024-09-11
fix: allow `-` in hostnames for Netlify Image CDN

hrishikesh-k:
## Changes

Fix for: https://answers.netlify.com/t/124975/

TL;DR: The current RegEx did not match `a-us.storyblok.com` as the hostname as the RegEx only allowed lowercase letters and numbers. This PR adds `-` to the list of allowed characters.

Note that, the RegEx is not very strict. For example, as per the URL spec, the `-` should not be as the first or last character in the hostname, but this RegEx doesn't check that. It simply allows the existence of `-` in the hostname.

## Testing

There's (probably) nothing much to test here.

## Docs

N/A

### DIFF
--- a/.changeset/orange-countries-scream.md
+++ b/.changeset/orange-countries-scream.md
@@ -2,4 +2,4 @@
 '@astrojs/netlify': patch
 ---
 
-allow `-` in hostnames for Netlify Image CDN RegEx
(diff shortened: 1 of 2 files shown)


ascorbic (approved):
Thanks!

PR #392, opened 2024-09-13
feat: update adapters supported features to account for new shape for Astro 5

Princesseuh:
## Changes

In Astro 5, the section on assets has been removed from supportedAstroFeatures because it made no sense, you cannot not support "assets" as a concept, otherwise your adapter either can't have routes (it doesn't support having an endpoint), or does not support hosting assets in general (so you don't support any imports that results in a file, no JS, no CSS, no images etc)

In truth, adapters always support `astro:assets` the feature, what they don't support is the default image service that relies on Node API at runtime. As such, options we updated to reflect only that.

## Testing

N/A

## Docs

N/A

### DIFF
--- a/packages/netlify/src/index.ts
+++ b/packages/netlify/src/index.ts
@@ -455,12 +455,7 @@ export default function netlifyIntegration(
 						hybridOutput: 'stable',
 						staticOutput: 'stable',
 						serverOutput: 'stable',
-						assets: {
-							// keeping this as experimental at least until Netlify Image CDN is out of beta
-							supportKind: 'experimental',
-							// still using Netlify Image CDN instead
-							isSharpCompatible: true,
-						},
+						sharpImageService: 'stable',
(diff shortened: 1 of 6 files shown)


packages/netlify/src/index.ts
```diff
@@ -455,12 +455,7 @@ export default function netlifyIntegration(
 						hybridOutput: 'stable',
 						staticOutput: 'stable',
 						serverOutput: 'stable',
-						assets: {
-							// keeping this as experimental at least until Netlify Image CDN is out of beta
-							supportKind: 'experimental',
```
Princesseuh, line 460: It makes no sense for Netlify to use this to reflect the Image CDN service being experimental. This leads to a wonky spooky messages for users (assets are not supported, what does that mean?)

florian-lefebvre (approved):
LGTM! Do we need a changeset?

PR #394, opened 2024-09-16
Reduce amount of rules in routes json file

veitbjarsch:
## Changes
Cloudflare has a limit of 100 includes/exclude rules. In many applications this amount is reached quite fast.
This PR add the functionality to compare include and exclude rules and reduce the amount of rules wherever possible.
Therefore it:
- adds a wildcard if a folder only belogs to include or exclude rules
- skips all subfolders where a wildcard is already present (leave them untouched)

## Testing

Tested in out own project using cloudflare as well wrote a unit test here.

### DIFF
--- a/.changeset/twenty-deers-wink.md
+++ b/.changeset/twenty-deers-wink.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Added functionality to compare include and exclude rules to reduce the amount of cloudflare rules
--- a/.changeset/weak-penguins-explode.md
+++ b/.changeset/weak-penguins-explode.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixed a bug where static file path was not split correctly
--- a/.changeset/weak-penguins-explode.md
+++ b/.changeset/weak-penguins-explode.md
@@ -1,5 +1,0 @@
----
-'@astrojs/cloudflare': patch
----
-
-Fixed a bug where static file path was not split correctly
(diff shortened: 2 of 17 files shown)


alexanderniebuhr:
> adds a wildcard if a folder only belogs to include or exclude rules

I have to check this internally, but I think we can't tell this upfront, either in all cases or in specific cases.

> skips all subfolders where a wildcard is already present (leave them untouched)

That's a good addition, we need to see how it handles the routes if you have all static in one folder and then one dynamic route in a nested subfolder. I'll take a look at your testcase.

veitbjarsch:
@alexanderniebuhr Thanks for your fast response.

>> skips all subfolders where a wildcard is already present (leave them untouched)
>
> That's a good addition, we need to see how it handles the routes if you have all static in one folder and then one dynamic route in a nested subfolder. I'll take a look at your testcase.

The code compares both includes and excludes in 2 ways. This means if the same folder (partial path) is present in both PathTrie objects, it will leave them untouched. Only if a folder is not present in the other PathTrie object we can savely assume that all children belongs to the current PathTrie. Only in this case we add the wildcard flag for this folder.

This way we can reduce the amount of rules in our project by more than 50% automatically, without the need to create and handle the routes.jsnon manually.

alexanderniebuhr:
Sounds good, I did a quick test and it seems to be working with all the cases. I'll give it a proper review later :)

veitbjarsch:
Thanks

alexanderniebuhr:
@veitbjarsch can you add an changeset with `pnpm exec changeset`

veitbjarsch:
@alexanderniebuhr added the changeset.

alexanderniebuhr (approved):

veitbjarsch:
@alexanderniebuhr Hey, is it planned to merge this PR into the main branch?

PR #395, opened 2024-09-17
chore: fix release notification step

alexanderniebuhr:
## Changes

- this should fix the notification script

## Testing

- existing

## Docs

- not needed

### DIFF
--- a/scripts/notify/index.js
+++ b/scripts/notify/index.js
@@ -1,6 +1,6 @@
 import { readFile } from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
-import { globby as glob } from 'globby';
+import glob from 'tiny-glob';
 
 const { GITHUB_REF = 'main' } = process.env;
 const baseUrl = new URL(`https://github.com/withastro/adapters/blob/${GITHUB_REF}/`);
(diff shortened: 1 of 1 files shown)


scripts/notify/index.js
```diff
@@ -82,7 +82,7 @@ function singularlize(text) {
 const packageMap = new Map();
 async function generatePackageMap() {
 	const packageRoot = new URL('../../packages/', import.meta.url);
-	const packages = await glob(['*/package.json', '*/*/package.json'], {
+	const packages = await glob('*/package.json', {
```
alexanderniebuhr, line 85: Looking at the current structure from the repo, this glob should be enough, I don't think we have two level nested packages 🤔

delucis (approved):
Seems plausible! 👍

scripts/notify/index.js
```diff
@@ -1,6 +1,6 @@
 import { readFile } from 'node:fs/promises';
 import { fileURLToPath } from 'node:url';
-import { globby as glob } from 'globby';
+import glob from 'tiny-glob';
```
ematipico, line 3: I would suggest using https://www.npmjs.com/package/tinyglobby, which is even smaller

bluwy (approved):
LGTM. I think we can switch to tinyglobby altogether later (since we're using a mix of tiny-glob and fast-glob currently)

alexanderniebuhr:
I probably should have made this PR against `main` instead of `next` 🤔

bluwy:
I think you can change it from the edit UI directly if you'd like

scripts/notify/index.js
alexanderniebuhr: As said in [this comment](https://github.com/withastro/adapters/pull/395#pullrequestreview-2309960412) let's switch altogether in a follow-up :)

PR #401, opened 2024-09-20
fix(netlify): correctly handle local images with fully-qualified URLs

ascorbic:
## Changes

Netlify Image CDN expects local images to just have the path. If passed the full URL it will need to handle domain allowlists, and it breaks locally. This PR detects those cases where an image URL is actually local and changes them to just use the path.

## Testing

Added test cases


packages/netlify/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.base.json",
   "include": ["src"],
   "compilerOptions": {
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "types": ["vite/client"]
```
bluwy, line 6: IIRC setting types here would make `@types/*` not work, could we use this comment instead on the top of the file that uses `import.meta.env.*`?

```js
/// <reference types="vite/client" />
```

packages/netlify/src/image-service.ts
```diff
@@ -17,9 +17,18 @@ const service: ExternalImageService = {
...
 		const query = new URLSearchParams();
 
-		const fileSrc = isESMImportedImage(options.src)
+		let fileSrc = isESMImportedImage(options.src)
 			? removeLeadingForwardSlash(options.src.src)
 			: options.src;
+		const site = import.meta.env.SITE;
+		if (URL.canParse(fileSrc)) {
+			const url = new URL(fileSrc);
+			if (import.meta.env.NETLIFY_LOCAL && url.hostname === 'localhost') {
```
ematipico, line 26: Would this work if the user starts the astro dev server with `--host`? This option exposes the server to the local network, so `localhost` isn't valid anymore

packages/netlify/src/image-service.ts
```diff
@@ -17,9 +17,18 @@ const service: ExternalImageService = {
...
+				fileSrc = fileSrc.substring(url.origin.length);
+			} else if (site && fileSrc.startsWith(site)) {
+				fileSrc = fileSrc.substring(site.length);
+			}
+		}
```
bluwy, line 31: Do you have an example where it's plausible for someone to write image paths with the SITE (which is usually the prod site url), or `localhost`, which theoretically would've only worked locally?

If this is more of the URL being mismanaged and passed in this way, maybe it makes more sense to fix on the consumer side?

ascorbic: The examples I've seen use `new URL(src, Astro.url.origin).href`, and they do it because `inferSize` needs a full URL. Maybe we should allow inferSize to work with local paths too.

packages/netlify/src/image-service.ts
ascorbic: It only works if you're running it using the Netlify CLI (which is what the first env var check is), which wouldn't work with `--host`

ascorbic:
Closing because I'm not sure this is somethign we should be encouraging

PR #406, opened 2024-09-27
chore: improve support for Workers Assets beta

alexanderniebuhr:
## Changes

- this improves support for Workers Assets Beta, mainly getting rid of not needed Platform files and making sure the SSR bundle does not get exposed if called by the `/_worker.js/index.js` url.
- Workers Assets is still in beta, so this is also only Beta supported by us for now
- more info to find
  - https://developers.cloudflare.com/workers/static-assets/
  - https://developers.cloudflare.com/workers/frameworks/framework-guides/astro/
  - https://blog.cloudflare.com/builder-day-2024-announcements/ 
- it should get paired with the following minimal `wrangler.toml` (not sure how we can provide that to the user)
```toml
name = ""
compatibility_date = "2024-09-19"
main = "./dist/_worker.js/index.js"
assets = { directory = "./dist/assets", binding = "ASSETS" }
```
## Testing

- not needed

## Docs

- needs a docs PR, will prepare one

cc @dario-piotrowicz @petebacondarwin


alexanderniebuhr:
I agree with the changeset, however I'm not sure about the test, because this is still `beta`. So I don't want our tests to fail, just because Cloudflare changes something. I guess it would be better to have tests only added once it is `stable` in Cloudflare.

ematipico:
>  I guess it would be better to have tests only added once it is stable in Cloudflare.

Can you explain why our tests would fail? I don't see any new dependency added, so I probably miss some context. Nonetheless, I trust your judgment. 

If these  workers assets are beta, I suppose that even **out option** should be _experimental_.

alexanderniebuhr:
Fail is the wrong term here, sorry. I meant the tests would not represent the `stable` expectations anymore, if the platform assumptions change.

But I agree the flag should be experimental. I'll change the code.

packages/cloudflare/src/index.ts
```diff
@@ -72,6 +71,17 @@ export type Options = {
...
+	/**
+	 * Lists all experimental features the adapter supports.
+	 */
+	experimental?: {
+		/**
+		 * Enables support for Cloudflare Workers assets. Defaults to false.
+		 */
+		cloudflare?: {
+			workerAssets?: boolean;
+		};
```
ematipico, line 83: What's the for a `cloudflare` field? The experimental option is **for the cloudflare** adapter, so I suppose there's no need for a `cloudflare` field. Is there something I miss?

alexanderniebuhr: `cloudflare` as platform features and not adapter or astro features?

ematipico: > platform/host feature and not adapter or astro feature

Oh, I see the point now. However, astro feature VS platform feature is an implementation detail, and it shouldn't be passed down to the user. Other adapters mix astro features and platform features options, which works well.

packages/cloudflare/src/index.ts
```diff
@@ -281,79 +296,81 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+							const parts = line.split(' ');
+							if (parts.length >= 2) {
+								const p = removeLeadingForwardSlash(parts[0])
+									.split('/')
+									.filter(Boolean)
+									.map((s: string) => {
+										const syntax = s
+											.replace(/\/:.*?(?=\/|$)/g, '/*')
+											// remove query params as they are not supported by cloudflare
+											.replace(/\?.*$/, '');
```
ematipico, line 327: Same here

alexanderniebuhr (commented):
Going to add some docs PR and going to address the nitpicks.

alexanderniebuhr (commented):
Going to add some docs PR and going to address the nitpicks.

packages/cloudflare/src/index.ts
```diff
@@ -281,79 +296,81 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
-							redirects.push(p);
+						for await (const line of rl) {
+							const parts = line.split(' ');
+							if (parts.length >= 2) {
+								const p = removeLeadingForwardSlash(parts[0])
+									.split('/')
+									.filter(Boolean)
+									.map((s: string) => {
+										const syntax = s
+											.replace(/\/:.*?(?=\/|$)/g, '/*')
```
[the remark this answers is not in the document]
alexanderniebuhr: Good point, I'll address it in a follow up PR, since it was like this before.

packages/cloudflare/src/index.ts
alexanderniebuhr: Good point, I'll address it in a follow up PR, since it was like this before.

alexanderniebuhr:
Actually I think it is just easier for users to add a ignore file, so we don't need to support two different build outputs.. closing

https://developers.cloudflare.com/workers/static-assets/binding/#ignoring-assets

Issue #407, opened 2024-09-29
Consider treating `cloudflare:sockets` as external in rollupOptions set in @astrojs/cloudflare


irvinebroque:
### Describe the Bug

When people try to build Astro apps that use the Postgres.js driver, they see this error at build time:

```
Error: Cannot resolve "cloudflare:sockets" from "...\node_modules\.pnpm\postgres@3.4.0\node_modules\postgres\cf\polyfills.js" and externals are not allowed!
```

https://x.com/masylum/status/1839743841252905467

Yes — people can fix this in userland by explicitly setting their vite config to treat `cloudflare:sockets` as external:

```
vite: {
  ssr: { external: ['node:buffer', 'node:events', 'cloudflare:sockets', 'node:stream' ] },
  build: { rollupOptions: { external: ['cloudflare:sockets'] } },
},
```

If I understand the adapter correctly, it _seems_ like there is an opportunity here to do this for people automatically, and reduce confusion and take this problem out of userspace?

https://github.com/withastro/adapters/blob/main/packages/cloudflare/src/index.ts#L246

### Participation

- [X] I am willing to submit a pull request for this issue.

PR #409, opened 2024-09-30
chore: externalize `cloudflare:sockets` imports

alexanderniebuhr:
**UPDATE:** I changed the approach to use a vite plugin to externalize the imports based on `resolveId`, this should be more flexible and work for all modes.

## Changes

- closes https://github.com/withastro/adapters/issues/407
- I still don't know if it would work with `astro dev`, without using `wrangler`, because in a local runtime the import wouldn't work, but at least the config wouldn't complain anymore.

## Testing

- existing tests

## Docs

- bugfix not needed

### DIFF
--- a/.changeset/good-rivers-add.md
+++ b/.changeset/good-rivers-add.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Improves default config to support `cloudflare:sockets` imports
--- a/.changeset/good-rivers-add.md
+++ b/.changeset/good-rivers-add.md
@@ -1,5 +1,5 @@
 ---
-'@astrojs/cloudflare': minor
+'@astrojs/cloudflare': patch
 ---
 
-Improves default config to support `cloudflare:sockets` imports
+Fixes an issue where `cloudflare:` scoped imports made the build fail. We externalize all imports with the `cloudflare:` scope by default now.
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -129,15 +129,17 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							// https://developers.cloudflare.com/pages/functions/module-support/
 							// Allows imports of '.wasm', '.bin', and '.txt' file types
 							cloudflareModulePlugin,
+							{
+								name: 'vite:cf-imports',
+								enforce: 'pre',
+								resolveId(source) {
+									if (source.startsWith("cloudflare:")) {
+										return { id: source, external: true };
+									}
+									return null;
+								},
+							}
 						],
-						ssr: {
-							external: ['cloudflare:sockets'],
-						},
-						build: {
-							rollupOptions: {
-								external: ['cloudflare:sockets'],
-							},
-						},
 					},
 					integrations: [astroWhen()],
 					image: setImageConfig(args?.imageService ?? 'compile', config.image, command, logger),
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -137,7 +137,7 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							rollupOptions: {
 								external: ['cloudflare:sockets'],
 							},
-						}
+						},
 					},
 					integrations: [astroWhen()],
 					image: setImageConfig(args?.imageService ?? 'compile', config.image, command, logger),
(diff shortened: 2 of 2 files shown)


.changeset/good-rivers-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Improves default config to support `cloudflare:sockets` imports
```
ematipico, line 5: nit: since it's a `minor`, maybe the changeset should contain more information e.g. what it means for the user and they can do now

ematipico (approved):

packages/cloudflare/src/index.ts
```diff
@@ -130,6 +130,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
 							// Allows imports of '.wasm', '.bin', and '.txt' file types
 							cloudflareModulePlugin,
 						],
+						ssr: {
+							external: ['cloudflare:sockets'],
```
irvinebroque, line 134:
```suggestion
							external: ['cloudflare:sockets', 'cloudflare:workers'],
```

https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/rpc/

packages/cloudflare/src/index.ts
```diff
@@ -130,6 +130,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+						},
+						build: {
+							rollupOptions: {
+								external: ['cloudflare:sockets'],
```
irvinebroque, line 138:
```suggestion
								external: ['cloudflare:sockets', 'cloudflare:workers'],
```

irvinebroque (commented):
Added one more that should be externalized, thanks for this @alexanderniebuhr !!!

packages/cloudflare/src/index.ts
```diff
@@ -130,6 +130,14 @@ export default function createIntegration(args?: Options): AstroIntegration {
...
+							},
+						}
```
petebacondarwin, line 140: Could this functionality be bundled into the `cloudflareModulePlugin`. In Wrangler we have a plugin that looks like this, and handles all `cloudflare:` prefixed imports:

```ts
export const cloudflareInternalPlugin: Plugin = {
	name: "Cloudflare internal imports plugin",
	setup(pluginBuild) {
		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, () => {
			return { external: true };
		});
	},
};
```

alexanderniebuhr: That might be even the better solution. :)

irvinebroque: What if we gave you an [unjs/unplugin](https://github.com/unjs/unplugin) that you could just rely on, where the plugin was the same configuration that Wrangler uses with esbuild?

hat tip Edmund from our team (can't tag from this GH repo) for suggesting.

alexanderniebuhr: Only think to consider here is that we don't use `esbuild`, but rather `vite`. Which under the hood uses `esbuild` & `rollup`, so any plugin needs to make sure that it works for both cases in `vite`

Fryuni: You can pass a regex to `external` so `/^cloudflare:` will externalize everything with the prefix

.changeset/good-rivers-add.md
Fryuni: Is it even a breaking change? I don't see how this would break any case that was working. It might change the error in some scenarios, but not add new ones.

packages/cloudflare/src/index.ts
alexanderniebuhr: I changed the PR to use this approach now.

alexanderniebuhr:
I changed the approach to use a vite plugin to externalize the imports based on resolveId, this should be more flexible and work for all modes, which follows [this suggestion](https://github.com/withastro/adapters/pull/409#discussion_r1782430390)

alexanderniebuhr:
I still don't think this is a breaking change or that we need to update the docs, because without this change, user had to add the code manually before.. `cloudflare:` imports can't be bundled, since those are runtime imports AFAIK. So before if the user did not externalize those imports manually, they ended up with an error during `astro build` 🤔

ematipico:
I see. Sounds like more a bugfix then.

Fryuni (commented):
Code seems great. Changeset needs update.

I agree with Ema that it's a bugfix.

ematipico (approved):

Fryuni (approved):

Issue #374, opened 2024-09-03
Cloudflare: _routes.json is truncated


schummar:
### Astro Info

```block
Astro                    v4.11.5
Node                     v20.12.2
System                   Linux (x64)
Package Manager          npm
Output                   hybrid
Adapter                  @astrojs/cloudflare
Integrations             auto-import
                         @astrojs/mdx
                         @astrojs/react
                         @astrojs/tailwind
                         @astrojs/sitemap
                         @playform/compress
```

### Describe the Bug

Hi, I have worked on automatic `_routes.json` generation some time ago. I tried to make the file as small as possible, because Cloudflare imposes a surprising limit of 100 rules. I guess there were issues or the feature or it was not possible to maintain with the big refactor or something. In any case, it's back to the old, simple way of generating the file.

Simple is good of course, but the way the limit is handled now is: After 100, the rest of the rules is simply cut off. I was very surprised that my page was not working as expected - because many of my static routes resulted in function invocations. That can be a lot slower and it can become really expensive, if one it not careful.

### What's the expected result?

Hi, I have worked on automatic `_routes.json` generation some time ago. I tried to make the file as small as possible, because Cloudflare imposes a surprising limit of 100 rules. I guess there were issues or the feature or it was not possible to maintain with the big refactor or something. In any case, it's back to the old, simple way of generating the file now.

Simple is good of course, but the way the limit is handled now is: After 100, the rest of the rules is simply cut off. I was very surprised that my page was not working as expected. A lot of static routes are resulting in function invocations instead! That can be a lot slower and expensive if you have a bit more traffic. I think, truncating is not the best solution. It is a common use case that sites have many static pages and only a few dynamic ones.

Would you be interested in investigating a more efficient approach once again? If so, what were the reasons/problems that prompted that simplification?

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/github-qvzkwv?file=dist%2F_routes.json

### Participation

- [X] I am willing to submit a pull request for this issue.

alexanderniebuhr:
Yeah we needed to rethink the implementation, due to some limitations. Could you explain what a more `efficient` approach is in your opinion?

schummar:
Sure. I realize there are probably some special cases to consider, but on the surface it could work like this - that's how I built it at the time:

1. A list of static routes and a list of dynamic routes is calculated - or I guess this info is already available somewhere
2. Try to merge entries in each list if they have common prefixes - but only as far this does not create a conflict with the other list
[...]

PR #423, opened 2024-10-14
feat: improve _routes.json generation

schummar:
## Changes

This is a result of the discussion in #374. I found a quick win that makes things a lot better without additional risk. More could be done to generate smaller, correct solutions for even more cases - but as a first step I'm pretty happy with this.

Problem: When there are a lot of static pages (resulting in more than 100 combined include/exclude entries, the adapter would generate a _routes.json with a `/*` include and a truncated array of excludes. Even if having a few `includes` and no `excludes` instead would be a correct solution too - with the added benefit of no unnecessary function invocations.

Solution: I improved the condition which switches between the two options (option 1 = `/*` include + many excludes; option 2 = many includes, fewer excludes). Now option 2 is chosen if: there is a static 404 AND option 2 is below the limit.

## Testing

I added two test cases with many static files. One with and one without a static 404 file.

## Docs

Not sure if anything needs to be documented.


.changeset/many-hornets-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+Better \_routes.json file for projects with many static pages
```
alexanderniebuhr, line 5:
```suggestion
Changes the logic which generates the `_routes.json` file to improve generation for projects with many static pages, while still making sure all routes work as expected.
```

alexanderniebuhr (approved):
This seems to make sense to me.

PR #424, opened 2024-10-18
feat(vercel): single entrypoint 

ematipico:
## Changes

Closes PLT-2550

This PR refactors the code of the Vercel adapter, and it now exports one single entry point for both static and serverless adapter.

The new code is inside `index.ts`. The code that differs between the new adapters is conditionally applied using `if (buidOutput === "server")` all over.

Also, I took the opportunity to make the creation of the `config.json` a bit more readable with a classic `if/else`, all those big ternaries were a bit hard to handle.

## Testing

The current tests should pass.

I only updated one single test, where Astro was throwing an error for an incorrect configuration. 

## Docs

https://github.com/withastro/docs/pull/9755


matthewp (approved):

florian-lefebvre:
Does this require changes in `astro add`?

ematipico:
> Does this require changes in `astro add`?

Good call. Yes, it will. I will make a PR

.changeset/fast-singers-design.md
```diff
@@ -0,0 +1,16 @@
+---
+'@astrojs/vercel': minor
+---
+
+The entrypoints `@astrojs/vercel/serverless` and `@astrojs/vercel/static` are deprecated. Use the `@astrojs/vercel` entrypoint
```
sarah11918, line 5:
```suggestion
Replaces the entrypoints `@astrojs/vercel/serverless` and `@astrojs/vercel/static` with a common `@astrojs/vercel` entrypoint.

You no longer specify an endpoint for your type of Vercel deployment, and you must update all your import statements to use the common entrypoint:
```
Would this be a correct way to say it? Or "Combines ... into"?

sarah11918 (commented):
Looks good! I was just wondering whether a word like "Replaces" or "combines" might be a be a more user-friendly word re: guiding them towards the changes they'll have to make (e.g. a find/replace)

.changeset/fast-singers-design.md
ematipico: Since this is a `minor`, I would like at least to tell the users that those entrypoints are *deprecated*, but they still work. I would like to avoid phrases like "replaced"/"removed" etc., because IMHO they hint a breaking change, which is not the case

packages/vercel/src/image/shared.ts
```diff
@@ -19,10 +19,10 @@ export type DevImageService = 'sharp' | (string & {});
...
 
 type RemotePattern = {
-	protocol?: 'http' | 'https';
-	hostname: string;
-	port?: string;
-	pathname?: string;
+	port?: string | undefined;
+	protocol?: string | undefined;
+	hostname?: string | undefined;
+	pathname?: string | undefined;
```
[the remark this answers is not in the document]
ematipico: Because the previous types were returning a type error. There is a type mismatch between Vercel ones and Astro ones.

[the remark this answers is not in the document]
bluwy: Could we force cast the Astro ones to the Vercel ones instead? I'm afraid we may lose this information about Vercel only allowing `http` or `https` if we remove this here, even if we don't check it in runtime now.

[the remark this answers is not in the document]
ematipico: Check if that's what you meant: [`1ae809f` (#424)](https://github.com/withastro/adapters/pull/424/commits/1ae809f647bfedb61a037625ee9d47a93c9f798d)

packages/vercel/src/index.ts
```diff
@@ -0,0 +1,620 @@
...
+							...imagesConfig,
+							domains: [...imagesConfig.domains, ..._config.image.domains],
+							remotePatterns: [
+								// ...(imagesConfig.remotePatterns ?? []),
+								// ..._config.image.remotePatterns,
+							],
+						};
+						if (imagesConfig.remotePatterns) {
+							images.remotePatterns?.push(...imagesConfig.remotePatterns);
+						}
```
bluwy, line 452: I think we can remove the comments here since they're unused. Or could combine both as:
```suggestion
							remotePatterns: [
								...(imagesConfig.remotePatterns ?? []),
							],
						};
```

packages/vercel/src/index.ts
```diff
@@ -0,0 +1,620 @@
...
+						const remotePatterns = _config.image.remotePatterns;
+						if (isAcceptedPattern(remotePatterns)) {
+							images.remotePatterns?.push(remotePatterns);
+						}
```
bluwy, line 456: Since `_config.image.remotePatterns` is an array, I think we need to filter the array here instead

bluwy (commented):
The change works for me 👍 We can also do type casting since that's easier if you want, but runtime checks also looks good. I noticed the code already had type casting like here: https://github.com/withastro/adapters/blob/16d688347ef98666fd21842b686a3b132f992d6f/packages/vercel/src/image/shared.ts#L7-L8

packages/vercel/src/index.ts
ematipico: I am unsure I understand `_config.image.remotePatterns` **isn't** an array. That's what TS is telling me

bluwy: It shows an array for me though 🤔  

image

---
image

ematipico: AH!! I completely ignored the `[]` at the end... -.-

packages/vercel/src/index.ts
```diff
@@ -0,0 +1,616 @@
...
+		},
+	};
+}
+
+function isAcceptedPattern(pattern: any): pattern is RemotePattern {
+	if (pattern?.port.statsWith('http') || pattern?.port.startsWith('https')) {
+		return true;
+	}
+	return false;
+}
```
bluwy, line 480: I believe we should be checking using `protocol` instead and whether it is `'http'` or `'https'`. `protocol` is also nullable so we shouldn't need to ensure that to exist here. The only property that we need to check to exist is `hostname`.

.changeset/fast-singers-design.md
sarah11918:
```suggestion
Deprecates the entrypoints `@astrojs/vercel/serverless` and `@astrojs/vercel/static`.  These will continue to work but are no longer documented and will be removed in a future version.  Update to the `@astrojs/vercel` entrypoint as soon as you are able:
```
How about something like this then?

packages/vercel/src/index.ts
ematipico: Good catch, thank you! Yes the check was meant for protocol

packages/vercel/src/index.ts
```diff
@@ -0,0 +1,619 @@
...
+			},
+		},
+	};
+}
+
+function isAcceptedPattern(pattern: any): pattern is RemotePattern {
+	if (!pattern?.hostname) {
+		return false;
+	}
+	if (pattern?.protocol?.statsWith('http') || pattern?.protocol?.startsWith('https')) {
```
bluwy, line 479: I think we can be stricter here and check `=== 'http'` and `=== 'https'` exactly. Also, I think this check still requires `RemotePattern` to have the `protocol` field, however the types showed that it's optional

packages/vercel/src/index.ts
```diff
@@ -0,0 +1,619 @@
...
+
+function isAcceptedPattern(pattern: any): pattern is RemotePattern {
+	if (!pattern?.hostname) {
+		return false;
+	}
+	if (pattern?.protocol === 'http' || pattern?.protocol === 'https') {
+		return true;
+	}
+	return false;
+}
```
bluwy, line 483:
```suggestion
function isAcceptedPattern(pattern: any): pattern is RemotePattern {
	if (pattern == null) {
		return false;
	}
	if (!pattern.hostname) {
		return false;
	}
	if (pattern.protocol && (pattern.protocol !== 'http' || pattern.protocol !== 'https')) {
		return false;
	}
	return true;
}
```

>  Also, I think this check still requires RemotePattern to have the protocol field, however the types showed that it's optional

My code above should address this as mentioned before. Can you please double check the logic 🙏

bluwy (approved):

Issue #430, opened 2024-10-31
`@astrojs/vercel` - AdapterSupportOutputMismatch


phenomen:
### Astro Info

```block
Astro                    v5.0.0-beta.6
Node                     v18.20.3
System                   Linux (x64)
Package Manager          npm
Output                   static
Adapter                  @astrojs/vercel
Integrations             none
```

### Describe the Bug

After upgrading to `"astro": "5.0.0-beta.6"` and `"@astrojs/vercel": "8.0.0-beta.3"` I get an error:

```
[AdapterSupportOutputMismatch] The `@astrojs/vercel` adapter is configured to output a static website, but the project contains server-rendered pages. Please install and configure the appropriate server adapter for your final deployment.
  Error reference:
    https://docs.astro.build/en/reference/errors/adapter-support-output-mismatch/
  Stack trace:
    at new AstroError (file:///home/projects/withastro-astro-fewhsc/node_modules/astro/dist/core/errors/errors.js:37:5)
    at setAdapter (file:///home/projects/withastro-astro-fewhsc/node_modules/astro/dist/integrations/hooks.js:265:45)
    at Module.runHookConfigDone (file:///home/projects/withastro-astro-fewhsc/node_modules/astro/dist/integrations/hooks.js:262:59)
    at async AstroBuilder.run (file:///home/projects/withastro-astro-fewhsc/node_modules/astro/dist/core/build/index.js:196:23)
    at async build (file:///home/projects/withastro-astro-fewhsc/node_modules/astro/dist/cli/build/index.js:44:3)
 ELIFECYCLE  Command failed with exit code 1.
```

From what I understand, `static` is a correct output if just a few pages/endpoints have `export const prerender = false;`

In the linked reproduction, run `npm run build` to repro the error.

### What's the expected result?

Successful build.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/withastro-astro-fewhsc?file=package.json

### Participation

- [ ] I am willing to submit a pull request for this issue.

bluwy:
Looks like we need to update the vercel adapter code here to use the value of `buildOutput` from the `astro:config:done` hook instead of from the `astro:config:setup` hook: https://github.com/withastro/adapters/blob/aed689d6af8f02cf059689f289b98881b5a4e335/packages/vercel/src/index.ts#L277-L289. That way `setAdapter` uses the right value that doesn't trigger the error.

[...]

PR #437, opened 2024-11-08
fix(vercel): output build directory

ematipico:
## Changes

Closes https://github.com/withastro/adapters/issues/430
Closes PLT-2625

My latest PR that unified the adapters had a regression where the output directory couldn't be determined when `astro:config:setup` was called.

The adapter uses *known directories* that Vercel uses to deploy the assets. These directories are set when Astro triggers `astro:config:setup`, and these directories depend on `buildOutput`. **However**, in Astro v5 this value isn't the final value inside `astro:config:setup`.

For this reason, the directories are computed at the end of the build, instead. Based on that, we copy the assets to the correct folder. 

## Testing

The current tests should still be valid

## Docs

N/A

### DIFF
--- a/packages/vercel/src/index.ts
+++ b/packages/vercel/src/index.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync } from 'node:fs';
+import { cpSync, existsSync, mkdirSync, readFileSync } from 'node:fs';
 import { basename } from 'node:path';
 import { pathToFileURL } from 'node:url';
 import { emptyDir, removeDir, writeJson } from '@astrojs/internal-helpers/fs';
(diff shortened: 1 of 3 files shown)


packages/vercel/src/index.ts
```diff
@@ -204,6 +206,29 @@ export default function vercelAdapter({
...
+						imageService,
+						imagesConfig,
+						command,
+						devImageService,
+						config.image
+					),
+				});
+			},
+			'astro:config:done': ({ setAdapter, config, logger }) => {
+				buildOutput = config.output;
```
bluwy, line 230: I think you can get the final `buildOutput` directly from the `astro:config:done` hook instead. [It returns `buildOutput`](https://github.com/withastro/astro/blob/d63d87dcaea85e84acb26b70cbd650d6e1eea6df/packages/astro/src/types/public/integrations.ts#L187)

matthewp (approved):

PR #453, opened 2024-11-28
Update cloudflare-module-loader.ts

rishi-raj-jain:
## Changes

After the build happens with the cloudflare adapter, one of my applications threw error as attached. But the preview ran fine using wrangler. I then noticed then it was failing to read from a file that did not exist in the `dist/_worker.js` directory. I think this happens because the Astro cleanup happened but it had the older resultant map.

image

### DIFF
--- a/.changeset/quick-weeks-clap.md
+++ b/.changeset/quick-weeks-clap.md
@@ -1,0 +1,5 @@
+---
+"@astrojs/cloudflare": patch
+---
+
+Skip replacing contents of the file if not found after cleanup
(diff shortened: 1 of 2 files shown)


alexanderniebuhr (commented):
This change looks okay to me, however I'm not sure if it doesn't hide the real issue.. It should never try to cleanup files, which don't exist anymore.. I'm concerned that we hide something here, which we should investigate in more detail.

ematipico:
Hi, we moved the cloudflare adapter in a new repository: https://github.com/withastro/astro/tree/main/packages/integrations/cloudflare

PR #458, opened 2024-12-02
fix(actions): skip `astro actions` requests in `trailingSlash = always`

darkmaga:
## Changes
Closes [#12568](https://github.com/withastro/adapters/issues/464)

Problem: node adapter doesnt pass `astro actions` request and always redirect this to path with the `/` in the end

image
image

Solution: I just update staticHandler in case `trailing = always` to skip `astro actions` requests

## Testing

Added two test cases 

## Docs

N/A

### DIFF
--- a/packages/node/test/trailing-slash.test.js
+++ b/packages/node/test/trailing-slash.test.js
@@ -81,6 +81,16 @@ describe('Trailing slash', () => {
 				assert.equal(res.status, 200);
 				assert.equal(css, 'h1 { color: red; }\n');
 			});
+
+			it('Does not add trailing slash to astro actions', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}/some-base/_actions/test`, {
+					method: 'POST',
+				});
+				const json = await res.json();
+
+				assert.equal(res.status, 200);
+				assert.equal(json[1], 'Hello from server!');
+			});
 		});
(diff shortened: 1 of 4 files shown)


packages/node/test/trailing-slash.test.js
```diff
@@ -81,6 +81,16 @@ describe('Trailing slash', () => {
 				assert.equal(res.status, 200);
 				assert.equal(css, 'h1 { color: red; }\n');
 			});
+
+			it('Does not add trailing slash to astro actions', async () => {
+				const res = await fetch(`http://${server.host}:${server.port}/some-base/_actions/test`, {
+					method: 'POST',
+				});
```
ematipico, line 88: I am not sure I understand the test. The current use case is built to have `trailingSlash: always`, this means that calling routes **without slash** should be forbidden, hence return a `404` 🤔

ematipico: Probably yes. Not sure why, but I thought we had fixed the trailing slash issue.

darkmaga: Then i'll close this pr and gonna work with `action` implementation

> Probably yes. Not sure why, but I thought we had fixed the trailing slash issue.

Maybe because of this [pr](https://github.com/withastro/astro/pull/12280)

PR #469, opened 2024-12-05
Sessions

ascorbic:
## Changes

- What does this change?
- Be short and concise. Bullet points can help!
- Before/after screenshots can help as well.
- Don't forget a changeset! `pnpm exec changeset`


alexanderniebuhr:
@ascorbic let me know if I can help adding a default driver for the Cloudflare adapter once Sessions get's stable

ascorbic:
@alexanderniebuhr it would be great if you could help get Cloudflare support working. Would the user need to enable KV themselves, or could we do it automatically?

alexanderniebuhr:
@ascorbic happy to help getting this to work with Cloudflare. 

> Would the user need to enable KV themselves, or could we do it automatically?

This depends a bit on what we need. I don't think we can add a KV Namespace for the user, unless we would have an API Token of them. However we can definitely streamline the needed configuration, so the user only has to provide an binding id after they created the namespace in the Dashboard.

Where can I take a look at what is needed, so I can try to figure it out for Cloudflare?

ascorbic:
@alexanderniebuhr the config info is [here](https://unstorage.unjs.io/drivers/cloudflare). Users can already configure the cloudflare drivers themselves, so it would be worth trying that yourself. The goal would be to see how much of that we could do for them. Right now the way to manually do it would be something like:

```ts
experimental: {
  session: {
    driver: "cloudflare-kv-binding",
    options: {
      binding: "SESSIONS"
    }
  }
}
```

ascorbic:
Superseded by https://github.com/withastro/astro/pull/13145

PR #481, opened 2024-12-13
fix(netlify): don't directly assign locals

ascorbic:
## Changes

Astro 5 changes the rules to prevent middleware from reassigning locals in middleware. This broke edge middleware on Netlify, because it does that to assign the Netlify context.

## Testing

Tested locally and on Netlify

### DIFF
--- a/.changeset/great-ties-relate.md
+++ b/.changeset/great-ties-relate.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/netlify': patch
+---
+
+Fixes an error where edge middleware would incorrectly assign locals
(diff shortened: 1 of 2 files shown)


ematipico (approved):

florian-lefebvre (approved):
Can you check if we need to fix something similar for vercel?

ascorbic:
@florian-lefebvre no, it looks like that's ok: 

https://github.com/withastro/adapters/blob/ntl-edge-locals/packages/vercel/src/serverless/middleware.ts#L101

dmayo2:
I just experienced this issue. Would it be weird that the issue happened after many builds, just suddenly? My site was working fine then poof. crashing the edge function.
https://answers.netlify.com/t/edge-function-crashing/134131

Issue #410, opened 2024-10-02
Incorrect svelte bundle imported when using svelte 5 with CloudFlare adapter


giraugh:
### Astro Info

```block
Astro                    v4.15.10
Node                     v20.10.0
System                   macOS (arm64)
Package Manager          pnpm
Output                   server
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/svelte
```

### Describe the Bug

When running and accessing a built astro app that uses svelte components and the CloudFlare workers adapter, it emits `lifecycle_outisde_component` errors. Only applies when using svelte 5.

```
[wrangler:inf] Ready on http://localhost:8788
[wrangler:inf] GET / 200 OK (16ms)
✘ [ERROR] Uncaught (in promise) Error: lifecycle_outside_component

✘ [ERROR] Uncaught (in response) Error: lifecycle_outside_component
```

From a quick investigation, it seems like the components are importing the client bundle when running in SSR. If I import and log out svelte's `onMount()` method from the component, it emits the client implementation rather than the noop implementation expected on the server. Could relate to worker import conditions?

## Details

- This issue exists both when using the provided version of `vite-plugin-svelte` (v3) and overriding the version to v4.
- Issue does not seem to occur when using svelte 4
- Issue does not seem to occur when using a different astro adapter (e.g node)

### What's the expected result?

Expected page to render as normal without errors.

### Link to Minimal Reproducible Example

https://github.com/giraugh/astro-svelte-5-cloudflare-repro

### Participation

- [X] I am willing to submit a pull request for this issue.

muuvmuuv:
Rollback to this works for now:
- @astrojs/cloudflare: 11.0.4
- astro: 4.16.0

yuto343:
This seems to have reappeared in astro@5.0.9 and @astrojs/cloudflare@12.1.0 🥲 I downgraded to 12.0.1 to work around it.

PR #487, opened 2024-12-16
fix(cloudflare): resolve conditions for svelte

### LINKED ISSUE #485: [Astro + Svelte + Cloudflare] Cannot build Svelte component that contains $onMount() with Cloudflare adapter
### Astro Info

```block
Astro                    v5.0.5
Node                     v18.20.3
Package Manager          unknown
Adapter                  @astrojs/cloudflare
Integrations             @astrojs/svelte
```

### Describe the Bug

If I try to build the Astro website that contains Svelte component with $onMount() and Cloudflare adapter, the build error will occurs like this.

Build error:

```
$ npm run build

prerendering static routes 
01:36:03 ▶ src/pages/index.astro
01:36:03   └─ /index.htmlhttps://svelte.dev/e/lifecycle_outside_component
  Stack trace:
    at lifecycle_outside_component (file:///home/projects/github-memgm1vc/dist/_worker.js/pages/index.astro.mjs?time=1734280563043:32:9)
    at Test (file:///home/projects/github-memgm1vc/dist/_worker.js/pages/index.astro.mjs?time=1734280563043:69:9)
    at renderToStaticMarkup (file:///home/projects/github-memgm1vc/dist/_worker.js/chunks/_@astro-renderers_BFYr5CxZ.mjs:1251:23)
```

It doesn't occurs when I use NodeJS adapter. It seems it is occurring after `@astrojs/cloudflare@12.1.0` update.

No build errors.

https://stackblitz.com/edit/astro-svelte-cloudflare-build-error

### Participation

rognstadragnar:
## Changes

Fixes #485.

#476 added the `workerd` and `worker` resolve conditions to fix the Vue build breaking. As this fix used the `defaultClientConditions` as a base it broke builds using Svelte. The client default conditions include the `browser` condition, which makes Svelte use the client entry when server side rendering. 

I feel that the `defaultServerConditions` make the most sense as a default here, and with filtering out the `node` condition the Vue build will still work.

## Testing

Added usage of `setContext` and `onMount` to the Cloudflare Svelte fixture. 
Without this fix the tests then fail. With the fix all tests pass.

## Docs

Should not really affect anything as this is just a bug fix.

### DIFF
--- a/.changeset/gorgeous-grapes-occur.md
+++ b/.changeset/gorgeous-grapes-occur.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': patch
+---
+
+Fixes a bug where svelte were not working while using the Cloudflare adapter
--- a/.changeset/gorgeous-grapes-occur.md
+++ b/.changeset/gorgeous-grapes-occur.md
@@ -2,4 +2,4 @@
 '@astrojs/cloudflare': patch
 ---
 
-Fixes a bug where svelte were not working while using the Cloudflare adapter
+Fixes a bug where Svelte was not working properly when using the Cloudflare adapter
--- a/packages/cloudflare/src/index.ts
+++ b/packages/cloudflare/src/index.ts
@@ -12,7 +12,7 @@ import {
 import { createRedirectsFromAstroRoutes } from '@astrojs/underscore-redirects';
 import astroWhen from '@inox-tools/astro-when';
 import { AstroError } from 'astro/errors';
-import { defaultClientConditions } from 'vite';
+import { defaultServerConditions } from 'vite';
 import { type GetPlatformProxyOptions, getPlatformProxy } from 'wrangler';
 import {
 	type CloudflareModulePluginExtra,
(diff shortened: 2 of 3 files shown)


ematipico (approved):

bluwy (changes requested):
I guess with #476, https://github.com/withastro/adapters/issues/410 resurfaced again and hasn't really been properly fixed. However, the behaviour now returns to the same as Vite 5 (and the previous major of the cloudflare adapter).

However, I don't know if this is the correct fix. We can't remove both "browser" and "node" conditions, either one needs to be enabled. I think I have to think a bit more about how to fix this first. There's an inconsistency where we're trying to use resolve conditions for Cloudflare (webworkers), but during dev, we're actually still running in node.

Blocking this for now but you can pin to the previous version that worked to workaround this at the meantime.

alexanderniebuhr:
@bluwy I feel like this is a good enough workaround for now. Are there any hard reasons not to merge this?

bluwy:
Closing this as fixed by https://github.com/sveltejs/svelte/pull/14779 (https://github.com/withastro/adapters/issues/485#issuecomment-2556402412)

PR #488, opened 2024-12-16
fix: correctly pass locals from Netlify edge middleware

### LINKED ISSUE #486: [netlify] [middleware] [edge] custom edge middleware are not setting locals (when running at edge)
### Astro Info

```block
Astro                    v5.0.5
Node                     v20.18.1
System                   Linux (arm64)
Package Manager          pnpm
Adapter                  @astrojs/netlify (patched with #481)
Integrations             none
```

### Describe the Bug

Following up [my previous issue](https://github.com/withastro/adapters/issues/483) that was closed by https://github.com/withastro/adapters/pull/481
I did try to patch the exact same fix you suggest in #481 before even raising the issue, but I did because this fix is not fixing the middleware fully. But wasn't super clear so let's move on step by step and here is ne next step investigation

TL;DR:

#481 is not fixing the issue, just avoiding a runtime error.
Here is the correct fix:

```diff
const ctx = createContext({
    request,
    params: {},
+   locals: { netlify: {context} },
});
- ctx.locals.netlify = { context }
```

*(Never contributed and took me already a lot of time investigating that so I might create the PR if you ask, but I'll do it tomorrow)*

---
> [!NOTE]
> The following is just the history of my investigation just for reference. You can skip reading !

## Tests

| Config Label | Working |
|--------------|---------|
| static + edge | ❌ 404 Server-Islands |
| static + not edge | ❌ 404 Server-Islands |
| server + edge | ❌ Missing Custom Local from middleware |
| server + not edge | ✅ Working |

### Test Result Analysis

* When a server-island component (ssr) is rendered inside a <ins>static page</ins> (ssg), the server-island endpoint is not created (404)
* When a server-island component (ssr) is rendered inside a <ins>dynamic</ins> page (ssr), the server-island works but the Astro.locals set by a custom `src/middleware.ts` file running <ins>at edge</ins> are ignored (the code is run from edge but has no impact on the response)
* When a server-island component (ssr) is rendered inside a <ins>dynamic</ins> page (ssr), <ins>not at edge</ins> everything works.

### Test Conclusion

it looks like there are two distinct problems. 
* One regarding the server-islands rendered inside a static page/component (`export const prerender = false;`)
* One regarding the custom astro middleware (`src/middleware.ts`) when running at edge on netlify (`edgeMiddleware: true`)

## Issue 1: Server island endpoint not building for static page

After further investigation, this appears only when all files in `src/pages` are static. <ins>And this is not a problem coming from netlify adapter</ins> as the output is `static` from `astro:config:done` hook.

> [!IMPORTANT]
> this must be a little edge case from Astro v5 when they merged the `static` and `hybrid` mode. Now it doesn't assume we are `hybrid` when we don't have any ssr page (even if they contains server-islands)

*I might open an issue on Astro repo if it's not yet beeing fixed...*

UPDATE: I did here: <https://github.com/withastro/astro/issues/12744>

> [!TIP]
> as a workaround simply add a unused page with `export const prerender = false;` and it will trick Astro to think we are no 100% static

## Issue 2: Locals from custom middleware are ignored when

If we apply my workaround to fix issue 1, here is the update test table for issue 2

| Config Label | Working |
|--------------|---------|
| static + edge | ❌ Missing Custom Local from middleware |
| static + not edge | ✅ Working |
| server + edge | ❌ Missing Custom Local from middleware |
| server + not edge | ✅ Working |

So the root cause is only coming from whether the middleware is run from/built for netlify edge. 
So I guess we still have an issue here regarding Netlify edge and custom middleware where 

> [!NOTE]
> The issue is only affecting the Astro.locals assignation. The custom headers written from custom middleware are working as expected !

After some further debug it looks like it comes exactly from the original issue caused by the context.locals api changes addressed in #481 but it does not fix it properly. The correct way is more this:

```diff
const ctx = createContext({
    request,
    params: {},
+   locals: { netlify: {context} },
});
- ctx.locals.netlify = { context }
```

whether we are in edgeMiddleware mode or not, the custom locals from custom `src/middleware.ts` should be passed properly to ssr pages and components.

https://github.com/PaulSenon/issue-reproduction-astro-netlify-adapter

### Participation

ascorbic:
## Changes

https://github.com/withastro/astro/pull/12647 changed the way locals need to be passed to middleware. This PR updates the Netlify adapter to use this.

Fixes #486

## Testing
Manually tested, and added test


packages/netlify/src/index.ts
```diff
@@ -289,9 +289,10 @@ export default function netlifyIntegration(
 			export default async (request, context) => {
 				const ctx = createContext({
 					request,
-					params: {}
+					params: {},
+					locals: { netlify: { context } }
 				});
-				ctx.locals.netlify = { context } 
+				console.log(ctx.locals)
```
ematipico, line 295: to remove

packages/netlify/src/index.ts
```diff
@@ -313,6 +314,7 @@ export default function netlifyIntegration(
 				};
 				const next = () => {
 					const { netlify, ...otherLocals } = ctx.locals;
+					console.log({otherLocals})
```
ematipico, line 317: remove

ematipico (approved):

PR #508, opened 2025-01-17
fix(vercel): copy static assets after all integrations

### LINKED ISSUE #445: @astrojs/sitemap excluded from final static output with Vercel adapter
### Astro Info

```block
Astro                    v5.0.0
Node                     v22.11.0
Package Manager          pnpm
Adapter                  @astrojs/vercel
Integrations             @astrojs/mdx
                         @astrojs/sitemap
```

### If this issue only occurs in one browser, which browser is a problem?

_No response_

### Describe the Bug

- https://github.com/withastro/astro/issues/12437

```json
"@astrojs/sitemap": "^3.2.1",
"@astrojs/vercel": "^8.0.0",
"astro": "^5.0.0",
```
output is `static`

Using the versions above, the sitemap-index.xml and sitemap-0.xml files are created in `/dist` instead of `.vercel/output/static`, excluding them from the final vercel deployment.

image

`astro-v4` branch in the repro

```json
"@astrojs/sitemap": "^3.2.1",
"@astrojs/vercel": "^7.8.2",
"astro": "^4.16.13",
```
output is `hybrid`

No `/dist` folder is created at all after building, and the sitemap is properly generated in `.vercel/output/static`.

```sh
[build] Rearranging server assets...
[@astrojs/sitemap] `sitemap-index.xml` created at `.vercel/output/static`
```

https://github.com/dotnize/astro-sitemap-vercel-static

### Participation

ascorbic:
## Changes

Adapters need to run before all other integrations, which is enforced in builds. This causes a problem for the Vercel adapter, which copies static assets from the dist folder into the `.vercel/static` folder after the build in the `'astro:build:done'` hook. For integrations that also generate assets in `'astro:build:done'` (such as [sitemap](#445) and [partytown](https://github.com/withastro/astro/issues/12832)), this means that they are too late, and generate the assets after the adapter has copied them to the deploy dir.

This PR does a bit of a sneaky trick: it adds a second integration, which *isn't* an adapter, and so can run after all others. That has its own `'astro:build:done'` hook which does the copying.

Fixes #445 

## Testing
Added fixture and test suite


florian-lefebvre (approved):
LGTM. Only issue I can see is that another integration could be added after this one from within an integration:

- adapter adds integration X
- integration Y adds integration Z
- X runs (adapter assets copy)
- Z runs

So if Z adds some assets they will not be copied. This should be rare tho. Maybe a way to work around that is for the adapter to inject an integration that injects an integration:

- adapter adds integration X
- integration Y adds integration Z
- X adds integration W
- Z runs
- W runs (adapter assets copy)

This is not bullet proof of course

ascorbic:
But what if another integration injects an integration that injects an integration? Maybe we should inject an integration that injects an integration that injects an integration that injects an integration

ascorbic:
Yeah, I was joking. Maybe we should add `pre` and `post` support, like Vite plugins.

dotnize:
This fix only seems to work on fully-static projects. The issue is still present when you have a non-static/hybrid route with `export const prerender = false`

https://github.com/withastro/adapters/issues/445#issuecomment-2608868822

PR #514, opened 2025-01-22
fix: correctly handle wildcard hosts in preview

ascorbic:
## Changes

Currently the Node adapter doesn't correctly handle the `host` option in preview.

This PR updates the adapter to use the same host rules as the astro core preview server: 

- `false` or unset uses default of `localhost`
- `true` uses wildcard host (`0.0.0.0`), meaning listen on all available host interfaces
- `undefined` (only set by using the cli flag with no argument) uses wildcard
- any other value uses that as the host

Fixes https://github.com/withastro/astro/issues/13034

## Testing

Adds test suite

### DIFF
--- a/.changeset/big-islands-add.md
+++ b/.changeset/big-islands-add.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/node': patch
+---
+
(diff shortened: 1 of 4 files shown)


ematipico (approved):
Looks good!

PR #515, opened 2025-01-22
Update netlify adapter to integrate includeFiles and excludeFiles options

dfdez:
## Changes
When trying Netlify SSR with Astro, I wasn't able to get the `included_files` option to work. After investigating, I realized that the `includeFiles` and `excludeFiles` functionalities, similar to what Vercel provides, were necessary to include additional files during the build process. 

In this PR, I have implemented the integration to add these options: 
- Added `includeFiles` and `excludeFiles` options to the integration, mirroring Vercel's functionality.
  - **`includeFiles`**: Allows users to explicitly specify files or directories to be bundled with their function. Useful for avoiding missing files in production.
  - **`excludeFiles`**: Enables users to specify files or directories to be excluded from the deployment. Helps minimize bundle size.
- All code for these features has been adapted directly from Vercel's integration for consistency and reliability.

## Testing

- Manually tested `includeFiles` and `excludeFiles` with various configurations to ensure:
  - Files specified in `includeFiles` are bundled correctly.
  - Files and directories specified in `excludeFiles` are omitted as expected.
- Automated tests added to validate edge cases and typical usage scenarios.
- Confirmed the implementation behaves identically to Vercel's integration.

## Docs

- Updated [astro docs](https://github.com/withastro/docs/pull/10760) with documentation for `includeFiles` and `excludeFiles` on netlify adapter.
- Added links to Vercel's documentation for further reference: 
  - [includeFiles](https://docs.astro.build/en/guides/integrations-guide/vercel/#includefiles)
  - [excludeFiles](https://docs.astro.build/en/guides/integrations-guide/vercel/#excludefiles)


ematipico:
> Automated tests added to validate edge cases and typical usage scenarios.

It doesn't seem you added new assertions, and you didn't test the `excludeFiles`. Can you please do so?

ematipico (changes requested):
Can you please add a new changeset?

ascorbic:
I have [an old PR](#325) that does the same, but it also included changing the cwd at runtime, which in retropect should have been separate – and proved to be controversial – which is why I never merged it. Mine has now become outdated and conflicted, so having a new PR like this makes sense to me, but you might want to look at my PR to get some tests and changeset ideas.

dfdez:
Thanks @ascorbic for the PR was very useful ✨

After taking a look to https://github.com/withastro/adapters/pull/325/ I have made the following changes:
- Added `glob` functionality based on https://github.com/withastro/adapters/pull/325/
- Added tests based on https://github.com/withastro/adapters/pull/325/
- Added changelogs copied from https://github.com/withastro/adapters/pull/325/

packages/netlify/test/functions/fixtures/includes/src/pages/index.astro
```diff
@@ -0,0 +1,27 @@
...
+import { promises as fs } from 'fs';
+import { fileURLToPath } from 'url';
+import { dirname, join } from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const loadFile = Astro.url.searchParams.get('file');
+
+const file = await fs.readFile(join(__dirname, `../../../files/${loadFile}`), 'utf-8');
```
dfdez, line 11: This was the only way that I was able to properly test this. Not sure if you might have an alternative solution to handle this scenario...

But while thinking about this, I wondered if it would be good idea to add the final asset path to the `netlify context`?

This would allow us to do something like:
```js
const file = await fs.readFile(join(Astro.locals.netlify.context.root, `./files/${loadFile}`), 'utf-8');
```

ascorbic: This was why I added the process.chdir in my PR. I think one of those options would be good, but should be in a different PR as it would need separate discussion

.changeset/beige-insects-hang.md
```diff
@@ -0,0 +1,23 @@
...
+'@astrojs/netlify': minor
+---
+
+Adds `includedFiles` and `excludedFiles` configuration options. These allow extra files to be deployed in the SSR function bundle.
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However, sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
+
+The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the site root. Both options support glob patterns, so you can include/exclude multiple files at once.
+
+If you are specifying files using filesystem functions, resolve the path using `path.resolve()` or `process.cwd()`, which will give you the site root. At runtime, compiled source files will be in a different location and you cannot rely on relative file paths.
```
ascorbic, line 11: I don't think this will work in this PR, because it doesn't do chdir

packages/netlify/test/functions/fixtures/includes/astro.config.mjs
```diff
@@ -0,0 +1,14 @@
+import netlify from '@astrojs/netlify';
+import { defineConfig } from 'astro/config';
+
+export default defineConfig({
+  vite: {
+    assetsInclude: process.env.VITE_ASSETS_INCLUDE?.split(',') || [],
```
ascorbic, line 6: Rather than juggling env vars you can pass the actual config to `loadFixture` in each test

packages/netlify/test/functions/include-files.test.js
```diff
@@ -0,0 +1,172 @@
...
+
+		const root = new URL('./fixtures/includes/', import.meta.url);
+		const expectedCwd = new URL('.netlify/v1/functions/ssr/packages/netlify/', root);
+
+		const expectedAssetsInclude = ['./*.json'];
+		const excludedAssets = ['./files/exclude-asset.json'];
+
+		before(async () => {
+			process.env.VITE_ASSETS_INCLUDE = expectedAssetsInclude.join();
+			fixture = await loadFixture({ root });
```
ascorbic, line 21: Pass the config to `loadFixture` instead

.changeset/beige-insects-hang.md
```diff
@@ -0,0 +1,23 @@
+---
+'@astrojs/netlify': minor
+---
+
+Adds `includedFiles` and `excludedFiles` configuration options. These allow extra files to be deployed in the SSR function bundle.
+
+When an Astro site using `server` or `hybrid` rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However, sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
```
sarah11918, line 7:
```suggestion
When an Astro site using on-demand rendering is deployed to Netlify, the generated functions trace the server dependencies and include any that may be needed in SSR. However, sometimes you may want to include extra files that are not detected as dependencies, such as files that are loaded using `fs` functions. Also, you may sometimes want to specifically exclude dependencies that are bundled automatically. For example, you may have a Node module that includes a large binary.
```
Noting that `hybrid` does not exist as a "code" word anymore (and, I'm not sure we use it casually?)

.changeset/beige-insects-hang.md
```diff
@@ -0,0 +1,23 @@
...
+
+The `includedFiles` and `excludedFiles` options allow you specify these inclusions and exclusions as an array of file paths relative to the site root. Both options support glob patterns, so you can include/exclude multiple files at once.
```
sarah11918, line 9: I've made a note (in the Docs PR) that this content did not make it into the docs PR, but should!

.changeset/beige-insects-hang.md
sarah11918: I can't speak to the tech, but will note:
- this feels a bit too detailed for a changelog
- if someone needs an explanation like this to use the feature, it must be in docs itself

sarah11918 (commented):
Just some tiny comments on the changeset, and info that seems to be in this PR that did not make it into docs!

packages/netlify/test/functions/fixtures/includes/astro.config.mjs
dfdez: Done 👍🏽  -> [63ce3bb](https://github.com/withastro/adapters/pull/515/commits/63ce3bbe00670cb91af95f9c3122fc166d098a69)

.changeset/beige-insects-hang.md
dfdez: Changelogs update -> https://github.com/withastro/adapters/pull/515/commits/5557cc0f2f6336029e6459b34726b4c7142feb9f

dfdez:
@ascorbic I've made the requested changes to the tests, and as @sarah11918 requested, I've updated the changelog and made some updates to the docs as well. 👍🏽 

Let me know if you need anything else! Thanks!

ematipico (commented):
I am not really confident about this feature because I don't know the code. I let @ascorbic doing th review

ascorbic (approved):
Looks good! Once this is in, you might want to think about your proposal for exposing the fucntion root dir at runtime. If we were to do it, it would make sense to do the same for Vercel.

ematipico:
Hi, we moved the netlify adapter in a new repository: https://github.com/withastro/astro/tree/main/packages/integrations/netlify

ascorbic:
@dfdez sorry we didn't get this in before the move. It should be fine to redo it as-is on the main repo, aside from the slightly changed base path

dfdez:
No worries! I'll redo it on the main repo 👍🏽

PR #516, opened 2025-01-23
fix(vercel): handle integration static assets in non-static sites

ascorbic:
## Changes

#508 fixed the deployment of static assets that were generated by integrations. However it only worked on static sites, and wasn't working correctly for server-rendered sites. This PR fixes that.

## Testing
Adds extra tests for both render modes


delucis (approved):
Thanks @ascorbic! LGTM 👍

PR #521, opened 2025-01-24
feat(cloudflare): add redirect routes with trailing slash

mschoeffmann:
## Changes

- fixes withastro/astro#13165
- duplicates every route in `_redirects` with added trailing slash to avoid 404 by default
- offers adapter configuration flag `redirectAlsoWithTrailingSlash` to disable this behaviour

## Testing

No additional test added because there was no initial test targeting `_redirects' was found.

## Docs

Should be added to docs, including the `redirectAlsoWithTrailingSlash` option.

### DIFF
--- a/.changeset/curly-eels-help.md
+++ b/.changeset/curly-eels-help.md
@@ -1,0 +1,5 @@
+---
+'@astrojs/cloudflare': minor
+---
+
+add redirect routes with trailing slash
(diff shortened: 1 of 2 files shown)


ematipico:
@mschoeffmann can you create a **new test** that covers the use case of the original bug?

alexanderniebuhr:
I still wonder if this isn't a bug which should be addressed in https://github.com/withastro/astro/tree/main/packages/underscore-redirects 🤔

ematipico:
Hi, we moved the cloudflare adapter in a new repository: https://github.com/withastro/astro/tree/main/packages/integrations/cloudflare

